PR #2, opened 2018-07-27
MM-10657 Adding file upload button to demo plugin.

crspeller:
- Also fixed Makefile and vgo files just enough to get it to compile for someone who already has vgo installed. 

Webapp PR:  https://github.com/mattermost/mattermost-webapp/pull/1488

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-GO=$(shell go env GOPATH)/bin/vgo
+GO=env CGO_ENABLED=0 $(shell go env GOPATH)/bin/vgo
 
 # Check that vgo is installed. This won't be necessary once Go 1.11 is released, but it will still
 # be necessary to assert the Go version.
@@ -83,7 +83,7 @@ dist: apply \
 # deploy installs the plugin to a (development) server, using the API if appropriate environment
 # variables are defined, or copying the files directly to a sibling mattermost-server directory
 .PHONY: deploy
-deploy:
+deploy: dist
 ifneq ($(and $(MM_SERVICESETTINGS_SITEURL),$(MM_ADMIN_USERNAME),$(MM_ADMIN_PASSWORD)),)
 	@echo "Installing plugin via API"
 	http --print b --check-status $(MM_SERVICESETTINGS_SITEURL)/api/v4/users/me || ( \
--- a/build/manifest/go.mod
+++ b/build/manifest/go.mod
@@ -3,7 +3,7 @@ module github.com/mattermost/mattermost-plugin-demo/build/manifest
 require (
 	github.com/gorilla/websocket v1.2.0 // indirect
 	github.com/hashicorp/go-hclog v0.0.0-20180402200405-69ff559dc25f // indirect
-	github.com/mattermost/mattermost-server v0.0.0-20180718195328-2edc5d4fe051
+	github.com/mattermost/mattermost-server v5.1.0-rc1.0.20180727213543-6ac82d517176+incompatible
 	github.com/nicksnyder/go-i18n v1.10.0 // indirect
 	github.com/pborman/uuid v0.0.0-20180122190007-c65b2f87fee3 // indirect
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -83,7 +83,7 @@ dist: apply \
 # deploy installs the plugin to a (development) server, using the API if appropriate environment
 # variables are defined, or copying the files directly to a sibling mattermost-server directory
 .PHONY: deploy
-deploy:
+deploy: dist
```
lieut-data: I'd prefer we decoupled `deploy` and `dist`, since the two together take longer and I often test just `deploy` in isolation. It's not hard for someone to do a `make dist deploy` if they actually want that.

build/manifest/go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/mattermost/mattermost-plugin-demo/build/manifest
 require (
 	github.com/gorilla/websocket v1.2.0 // indirect
 	github.com/hashicorp/go-hclog v0.0.0-20180402200405-69ff559dc25f // indirect
-	github.com/mattermost/mattermost-server v0.0.0-20180718195328-2edc5d4fe051
+	github.com/mattermost/mattermost-server v5.1.0-rc1.0.20180727213543-6ac82d517176+incompatible
```
lieut-data: Can you share how you updated `go.mod`? It's odd that it would point at `5.1.0` instead of `master` with the actual latest manifest changes. Anyway, I have a set of changes to rip out `vgo` as discussed anyway, but more curious than anything.

Makefile
crspeller: In what situation would you want to deploy but not deploy the latest code? It would be really easy for someone while quickly iterating to do a `make deploy` without a `make dist` and deploy the wrong thing. This insures you always deploy the latest.

build/manifest/go.mod
crspeller: I just typed in `master` vgo changed it to that.

Makefile
lieut-data: Sometimes what I've got pending won't compile yet, but I'm still iterating on something that involves adding or removing the plugin. I've even done a `deploy` intentionally when the changes do build but I want to verify something on the previous build.

crspeller: Not sure I understand how that would help. Nothing changes when you deploy. You have to restart the server or enable/disable the plugin in the system console for any code to run.

lieut-data: There are [two modes](https://github.com/mattermost/mattermost-plugin-sample/blob/master/Makefile#L83) to `make deploy`:
* if `MM_SERVICESETTINGS_SITEURL`, `MM_ADMIN_USERNAME` and `MM_ADMIN_PASSWORD` are defined, [http](https://httpie.org/) is used to hit the API and remove any existing such plugin, install the new one, and enable it. Apart from having to run `make deploy` manually, this is a hot-loading-like experience, since not even the webapp must be refreshed.
* otherwise, it tries to copy the targets into the server folder as usual, and then requires your manual restart/enablement as necessary.

I use the former configuration exclusively, since it's so much faster for iteration.

crspeller: Ok. Not sure the benefits of that outweigh the issue with having it separate (getting confused about what version you actually deployed). If you want to iterate on the same plugin you can just click the enable/disable button in the system console, our write a little script to call the APIs to enable/disable. 4/5

Also that second mode I didn't take a close look at is cool. I will have to define those environment variables.

lieut-data: It's been a long time since I've seen a 4/5! I'm still of the mind that anyone who misunderstands the commands would immediately learn how they work and anyone who wishes they were separated would be otherwise annoyed, but I'm willing to give your workflow a go. It's possible my perspective is skewed to the plugin-framework-developer workflow vs. just the plugin-developer workflow :)

Re: the second mode, I note that we can eliminate that `http` dependency by just using the `Client4` API via something like the `manifest` binary we're already implementing. But I lifted it from local scripts I had already been using to do the same ;)

PR #6, opened 2018-09-19
Extract plugin version from manifest

Hanzei:
Downstream from https://github.com/mattermost/mattermost-plugin-sample/pull/8

This PR also bumps the plugins version to 0.0.2. Someone forgot to do this with https://github.com/mattermost/mattermost-plugin-demo/releases/tag/v0.0.2. It also updates the server dependencies to 5.3.2 and lock its there for better stability.


lieut-data:
@Hanzei, I've pushed a proposed PR to mattermost-plugin-sample (https://github.com/mattermost/mattermost-plugin-sample/pull/10) and wonder if I might get your feedback on same? If we merge the proposed changes, it would be great to update this PR as well.

Also, might you be able to submit the server dependencies update against mattermost-plugin-sample as well to keep things synchronized there?

lieut-data:
Lastly, while we're fixing the version number here, can we also bump the version to 0.0.3 given the mattermost-server dependency update and the potential for regressions? We'll tag a release accordingly :)

Hanzei:
@lieut-data I downstreamed your PR and bump the version. Please also note my last commit. I see no need for having non go files in local `vendor/`.

Hanzei:
> Also, might you be able to submit the server dependencies update against mattermost-plugin-sample as well to keep things synchronized there?

Sure

lieut-data:
Looks good to me! I agree re: non-go files :)

PR #8, opened 2018-10-02
Add demo usage of GetServerVersion 

hanzei:
This PR adds a demo usage of `GetServerVersion()' and add unit tests for this. It also changes some code to use Idiomatic error handling.

See https://github.com/mattermost/mattermost-server/pull/9429

### DIFF
--- a/server/Gopkg.toml
+++ b/server/Gopkg.toml
@@ -5,7 +5,7 @@
 
 [[constraint]]
   name = "github.com/mattermost/mattermost-server"
-  branch = "master"
+  branch = "release-5.4"
 
 [[constraint]]
   name = "github.com/blang/semver"
--- a/server/Gopkg.toml
+++ b/server/Gopkg.toml
@@ -5,4 +5,8 @@
 
 [[constraint]]
   name = "github.com/mattermost/mattermost-server"
-  version = "~5.3.0"
+  branch = "master"
+
+[[constraint]]
+  name = "github.com/blang/semver"
+  version = "3.5.1"
--- a/server/Gopkg.toml
+++ b/server/Gopkg.toml
@@ -9,4 +9,8 @@
 
 [[constraint]]
   name = "github.com/blang/semver"
-  version = "3.5.1"
+  version = "~3.5.0"
+
+[[constraint]]
+  name = "github.com/stretchr/testify"
+  version = "~1.2.0"
--- a/server/activate_hooks.go
+++ b/server/activate_hooks.go
@@ -13,7 +13,7 @@ const minimumServerVersion = "5.4.0"
 //
 // This demo implementation logs a message to the demo channel whenever the plugin is activated.
 func (p *Plugin) OnActivate() error {
-	v, err := semver.Parse(p.API.GetServerVersion())
+	serverVersion, err := semver.Parse(p.API.GetServerVersion())
 	if err != nil {
 		p.API.LogError(
 			"failed to parse server version",
@@ -22,21 +22,21 @@ func (p *Plugin) OnActivate() error {
 		return err
 	}
 	r := semver.MustParseRange(">=" + minimumServerVersion)
-	if !r(v) {
-		return fmt.Errorf("current Mattermost version is to low. Please update your Mattermost Server to a least v%s.", minimumServerVersion)
+	if !r(serverVersion) {
+		return fmt.Errorf("current Mattermost version is too low. Please update your Mattermost Server to at least v%s.", minimumServerVersion)
 	}
 
-	teams, appErr := p.API.GetTeams()
-	if appErr != nil {
+	teams, appError := p.API.GetTeams()
+	if appError != nil {
 		p.API.LogError(
 			"failed to query teams OnActivate",
-			"error", appErr.Error(),
+			"error", appError.Error(),
 		)
-		return appErr
+		return appError
 	}
 
 	for _, team := range teams {
-		_, appErr := p.API.CreatePost(&model.Post{
+		if _, appError := p.API.CreatePost(&model.Post{
 			UserId:    p.demoUserId,
 			ChannelId: p.demoChannelIds[team.Id],
 			Message: fmt.Sprintf(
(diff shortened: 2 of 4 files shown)


server/Gopkg.toml
```diff
@@ -5,4 +5,8 @@
 
 [[constraint]]
   name = "github.com/mattermost/mattermost-server"
-  version = "~5.3.0"
+  branch = "master"
```
hanzei: Once 5.4.0-rc1 is cut, we can change this.

server/activate_hooks.go
```diff
@@ -3,41 +3,59 @@ package main
...
+		p.API.LogError(
+			"failed to query teams OnActivate",
+			"error", appErr.Error(),
+		)
+		return appErr
 	}
 
 	for _, team := range teams {
-		if _, err := p.API.CreatePost(&model.Post{
+		_, appErr := p.API.CreatePost(&model.Post{
```
lieut-data: @hanzei, is this the idiomatic error handling changes to which you referred? The code as was written was already idiomatic. The changes we've been making in the main codebase and elsewhere are to fix blocks like this:
```go
if x, err := DoSomething(); err != nil {
    // error handling
    return err
} else {
    // do something with x

    if y, err := DoSomethingElse(); err != nil {
        // ...
    }
}
```

to avoid the nesting and return early:
```go
x, err := DoSomething(); 
if err != nil {
    // error handling
    return err
}

// do something with x

y, err := DoSomethingElse(); 
if err != nil {
        // ...
}
```

If we're not using the return value of the function outside the error handling block, it's fine to declare them inline in the conditional and handle immediately:
```go
if err := DoSomethingLast(); err != nil {
    // error handling
    return err
}
```

server/activate_hooks.go
```diff
@@ -3,41 +3,59 @@ package main
...
 )
 
+const minimumServerVersion = "5.4.0"
+
 // OnActivate is invoked when the plugin is activated.
 //
 // This demo implementation logs a message to the demo channel whenever the plugin is activated.
 func (p *Plugin) OnActivate() error {
-	teams, err := p.API.GetTeams()
+	v, err := semver.Parse(p.API.GetServerVersion())
```
lieut-data: nit: prefer clear names like `serverVersion` over single letter names like `v`.

server/activate_hooks.go
```diff
@@ -3,41 +3,59 @@ package main
...
 		p.API.LogError(
-			"failed to query teams OnActivate",
+			"failed to parse server version",
 			"error", err.Error(),
 		)
+		return err
+	}
+	r := semver.MustParseRange(">=" + minimumServerVersion)
+	if !r(v) {
+		return fmt.Errorf("current Mattermost version is to low. Please update your Mattermost Server to a least v%s.", minimumServerVersion)
```
lieut-data: In this particular case, we'll actually fail on line 17 above given that `GetServerVersion` won't be implemented prior to `5.4.0`. This seems like a good opportunity to teach a potential user of the API about this nuance. 

What if we change the behaviour here not to enforce a minimum, but to modify the `OnActivate` message posted to the channel below to dump the `model.Version` against which the plugin was compiled alongside the result from  `p.API.GetServerVersion()`, displaying `unknown` if the API isn't actually available as such (because an error was returned). We use this plugin during release testing, and I think it would be more useful to keep the plugin "working" across versions (with some useful version dumping to prove the API call is correct) than failing outright.

It might also be interesting to formalize the error returned when an API isn't implemented, but that's an idea for another PR I suppose :)

server/activate_hooks_test.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"testing"
+
+	"github.com/mattermost/mattermost-server/model"
+	"github.com/mattermost/mattermost-server/plugin/plugintest"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+)
+
+func TestOnActivate(t *testing.T) {
```
lieut-data: This is awesome! Great idea showcasing how to use the testing framework... and of course testing the plugin itself :D

server/Gopkg.toml
lieut-data: Could we just point this to the `release-5.4` branch instead?

hanzei: Sure

server/activate_hooks.go
hanzei: Thanks for the clarification! I reverted my changes.

server/activate_hooks_test.go
hanzei: If you think it help, I would also work on adding more test cases.

lieut-data: By all means!

server/activate_hooks.go
lieut-data: Unfortunately, it doesn't panic: just an error log at present, returning zero values if any. I'm assuming the semver would fail on attempting to parse the resulting empty string.

I do agree that there's a dual audience here, and that we should favour plugin developers. Let's stick with your proposal.

server/activate_hooks.go
```diff
@@ -3,37 +3,54 @@ package main
...
 		p.API.LogError(
-			"failed to query teams OnActivate",
+			"failed to parse server version",
 			"error", err.Error(),
 		)
+		return err
+	}
+	r := semver.MustParseRange(">=" + minimumServerVersion)
+	if !r(serverVersion) {
+		return fmt.Errorf("current Mattermost version is to low. Please update your Mattermost Server to a least v%s.", minimumServerVersion)
```
lieut-data: "too" low and "at" least

server/activate_hooks.go
```diff
@@ -3,37 +3,54 @@ package main
...
+	}
+
+	teams, appErr := p.API.GetTeams()
```
lieut-data: can you revert this `appErr` change as well? (just leave it named `err`)

lieut-data: Oh, my mistake! I didn't clue into the fact that you had already declared `err` as an `error`, in which case `appError` is perfectly fine. I thought it was just an artifact of the other refactoring not undone.

The medium-term plan is to move towards the `error` interface and only deal with `AppError` when we need the fields it exposes. Otherwise, it's more idiomatic to use `error` everywhere, to avoid the kind of pain you're seeing here.

hanzei:
I forgot to add `github.com/stretchr/testify/assert` to `Gopkg.toml`, woops. This is now fixed.

PR #9, opened 2018-10-04
Add tests for ServeHTTP and OnDeactivate

hanzei:
This PR adds tests for `OnDeactivate` and `ServeHTTP`.

### DIFF
--- a/server/activate_hooks_test.go
+++ b/server/activate_hooks_test.go
@@ -39,6 +39,29 @@ func TestOnActivate(t *testing.T) {
 			},
 			ShouldError: true,
 		},
+		"minimum supported version: 5.4.0, but CreatePost fails": {
+			SetupAPI: func() *plugintest.API {
+				api := &plugintest.API{}
+				api.On("GetServerVersion").Return("5.4.0")
+				api.On("GetTeams").Return([]*model.Team{&model.Team{Id: teamId}}, nil)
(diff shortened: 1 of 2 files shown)


hanzei:
I'm willing to write tests for more methods. Should they go into one PR or multiple ones like this?

hanzei:
@lieut-data ping

server/activate_hooks.go
```diff
@@ -54,7 +54,7 @@ func (p *Plugin) OnActivate() error {
 				"OnActivate: %s", manifest.Id,
 			),
 			Type: "custom_demo_plugin",
-			Props: map[string]interface{}{
+			Props: model.StringInterface{
```
lieut-data: 3/5, but I think we should stick with `map[string]interface{}` here instead of `model.StringInterface`. I think it's addition to `model` was premature: we don't add any methods to the type, and even methods like `StringInterfaceToJson` take a `map[string]interface{}`.

I suspect it will be less confusing for plugin authors to use the idiomatic type definition than an alias we use inconsistently.

lieut-data:
Adding @enahum for an additional set of eyes, and any extra context he might add :)

server/activate_hooks.go
hanzei: I get your point. Thanks for the review! I reverted my changes.

hanzei:
Changes made per review

PR #11, opened 2018-10-11
MM-11866 system-wide setting implementation example

drgsutu:
Adds system-wide setting implementation example.

Fixes https://github.com/mattermost/mattermost-server/issues/9378


jwilander:
Sorry for the late review here, this somehow slipped under my radar. This looks good for the most part but I think we need a way for the client-side of the plugin to get the `SystemWideSetting` on first load. Unless I'm mistaken, they will only get the setting as implemented if it changes while they are connected to the WebSocket, right?

hanzei:
@drgsutu Do you have some time to answer @jwilander questions above?
Thanks you very much for your contribution! :+1:

hanzei:
Closing this one due to inactivity. Still, thanks you for the contribution @drgsutu. :+1:

PR #13, opened 2018-10-17
Add check-style makefile target

hanzei:
Downstream from https://github.com/mattermost/mattermost-plugin-sample/pull/15


hanzei:
cc @lieut-data

lieut-data:
🎉

PR #19, opened 2018-12-26
Fixed OnDeactivate to UnregisterCommand (was Register...)

levb:


hanzei:
CI fails because you need to also change
- https://github.com/mattermost/mattermost-plugin-demo/blob/master/server/activate_hooks_test.go#L126
- https://github.com/mattermost/mattermost-plugin-demo/blob/master/server/activate_hooks_test.go#L156

levb:
@hanzei done. sorry I missed it, wrongly assumed `make dist` would run through the tests.

PR #20, opened 2019-02-13
MM-12765 Illustrate demo plugin settings schema

gabrieljackson:
This change enhances the demo plugin to include all current
setting options available to Mattermost plugins. Also, changes
in plugin settings are now announced via messages to the demo
channel just like other hooks.

This change also includes some code cleanup and additional tests
to increase code coverage.

Fixes https://github.com/mattermost/mattermost-server/issues/9695
https://mattermost.atlassian.net/browse/MM-12765


plugin.json
```diff
@@ -31,6 +31,64 @@
...
+            }, {
+                "display_name": "McDemo",
+                "value": "McDemo"               
+            }]
+        }, {
+            "key": "TextStyle",
+            "display_name": "Text Style",
+            "type": "dropdown",
+            "help_text": "Change the text style of the messages posted by this plugin",
+            "default": "none",
```
hanzei: The default value reference a value, not the display name.
 
```suggestion
            "default": "",
```

gabrieljackson: Oh, good catch. Sorry, that didn't manifest as an issue for me because I had changed the setting multiple times and kept redeploying the plugin.

This has been corrected.

gabrieljackson:
Sure thing! I will add that in shortly.

gabrieljackson:
I have added documentation to the server README for the new plugin settings. Also, I have made one small update to the webapp README to display the new config post type.

server/README.md
```diff
@@ -108,3 +109,47 @@ This demo implementation rejects login attempts by the demo user.
...
+
+![system-console-demo-settings](docs/demo-settings.png)
+
+## Channel Name
+
+A `text` setting type to define the channel that the demo plugin will create and use to post messages.
+
+## Username
+
+A `text` setting type to define the username value that will be given to the demo user. This is set via
```
jasonblais: `This is set via` accidentally left at the end of the line

server/README.md
```diff
@@ -108,3 +109,47 @@ This demo implementation rejects login attempts by the demo user.
 ### UserHasLoggedIn
 
 This demo implementation logs a message to the demo channel whenever a user logs in.
+
+# Plugin Settings
```
jasonblais: Given the README title uses heading 1 (with a single `#`), propose we use heading 2 (with `##`) for the Plugin Settings section, with the subsections using heading 3 (with `###`) like so:

```
## Plugin Settings

The following settings are available in the demo plugin system console page to demonstrate what is available via the [Mattermost Plugin Settings Schema](https://developers.mattermost.com/extend/plugins/manifest-reference/#settings_schema).

![system-console-demo-settings](docs/demo-settings.png)

### Channel Name

A `text` setting type to define the channel that the demo plugin will create and use to post messages.
```

server/README.md
gabrieljackson: Thanks for catching that

server/README.md
gabrieljackson: Updated

hanzei:
Thanks you very much for the contribution @gabrieljackson. This is amazing! :man_dancing:

PR #24, opened 2019-03-19
MM-14575: Automatically serve static files for plugins

happygaijin:
* Added sample public file
* Added HAS_PUBLIC to Makefile to test for public folder
* Added "static_files" config setting
* Related to [https://github.com/mattermost/mattermost-server/pull/10476](https://github.com/mattermost/mattermost-server/pull/10476l)

[https://github.com/mattermost/mattermost-server/issues/10447](https://github.com/mattermost/mattermost-server/issues/10447)


harshilsharma63:
@happygaijin any specific reason for keeping the `public` directory in `server` dir? I suppose it would be more flexible if it could be kept in either `server` or `webapp`, or maybe even both!

happygaijin:
> @happygaijin any specific reason for keeping the `public` directory in `server` dir? I suppose it would be more flexible if it could be kept in either `server` or `webapp`, or maybe even both!

I'm good with making that change. I was following a conversation in Developer Toolkit where they talked about that though. [Here is that conversation](https://community.mattermost.com/core/pl/rgnrq1k1aigpxbnh8sfwa6cmyr)

PR #25, opened 2019-04-05
MM-14860 Adding localization.

crspeller:
Implementation of: https://github.com/mattermost/mattermost-webapp/pull/2621

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,8 +155,10 @@ endif
 
 .PHONY: i18n-extract
 i18n-extract: ## Extract strings for translation from the source code
+ifneq ($(HAS_WEBAPP),)
 	@[[ -d $(MM_UTILITIES_DIR) ]] || echo "You must clone github.com/mattermost/mattermost-utilities repo in .. to use this command"
 	@[[ -d $(MM_UTILITIES_DIR) ]] && cd $(MM_UTILITIES_DIR) && npm install && npm run babel && node mmjstool/build/index.js i18n extract-webapp --webapp-dir ../mattermost-plugin-demo/webapp
+endif
 
 # clean removes all build artifacts
 .PHONY: clean
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ NPM ?= $(shell command -v npm 2> /dev/null)
 HTTP ?= $(shell command -v http 2> /dev/null)
 CURL ?= $(shell command -v curl 2> /dev/null)
 MANIFEST_FILE ?= plugin.json
+MM_UTILITIES_DIR ?= ../mattermost-utilities
 
 # Verify environment, and define PLUGIN_ID, PLUGIN_VERSION, HAS_SERVER and HAS_WEBAPP as needed.
 include build/setup.mk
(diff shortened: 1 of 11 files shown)


webapp/src/plugin.jsx
```diff
@@ -24,6 +30,16 @@ import {
 } from './actions';
 import reducer from './reducer';
 
+function getLanguageTest(locale) {
```
levb: why is function called getLanguage _Test_?

webapp/i18n/es.json
```diff
@@ -0,0 +1,13 @@
+{
```
hanzei: For my understanding: This file does **not** need to be in the bundle, correct?

webapp/src/plugin.jsx
crspeller: Because I forgot to change it.

webapp/src/components/left_sidebar_header/left_sidebar_header.jsx
```diff
@@ -25,7 +26,25 @@ export default class LeftSidebarHeader extends React.PureComponent {
...
+                <FormattedMessage
+                    id='sidebar.demo'
+                    defaultMessage='Demo Plugin:'
+                />
+                {' '}
+                {this.props.enabled ?
+                    <span>
+                        <FormattedMessage
+                            id='sidebar.enabled'
+                            defaultMessage='Enabled'
```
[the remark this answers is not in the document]
crspeller: This battle was fought long ago, defaultMessage is the way it is. It provides a default even if translations are missing, `en.json` is for the translations server. I don't mind now because the automation takes care of creating `en.json`.

Makefile
```diff
@@ -152,6 +153,11 @@ ifneq ($(HAS_WEBAPP),)
 	cd webapp && $(NPM) run fix;
 endif
 
+.PHONY: i18n-extract
+i18n-extract: ## Extract strings for translation from the source code
+	@[[ -d $(MM_UTILITIES_DIR) ]] || echo "You must clone github.com/mattermost/mattermost-utilities repo in .. to use this command"
```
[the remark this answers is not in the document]
crspeller: Fixed.

webapp/i18n/es.json
crspeller: Correct. It could be extracted using file-loader like in the webapp, but for simplicity I just imported it directly here. For most plugins this is probably fine as they won't have a lot of translations.

crspeller:
@hanzei Ready for re-review.

webapp/i18n/es.json
hanzei: I see. :+1: Importing it directly also seams the easier way here.

webapp/src/components/left_sidebar_header/left_sidebar_header.jsx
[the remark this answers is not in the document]
hanzei: Thanks for the explanation!

levb:
Is this ready to merge, or are we waiting on other PRs?

crspeller:
Waiting on the base PR.

PR #29, opened 2019-05-08
Use new bot helper.

crspeller:
Adds the bot helper from: https://github.com/mattermost/mattermost-server/pull/10542


server/configuration.go
```diff
@@ -186,6 +186,17 @@ func (p *Plugin) OnConfigurationChange() error {
 		return errors.Wrap(err, "failed to ensure demo user")
 	}
 
+	botId, ensureBotError := p.Helpers.EnsureBot(&model.Bot{
```
lieut-data: Happy to see `error`  in the `Helpers` interface, but I fear that without a holistic change to apply this across the API and Hooks, we're making it very hard on plugin authors who won't know of the need to distinguish `err` from `ensureBotError` in this case.

server/configuration.go
```diff
@@ -186,6 +186,17 @@ func (p *Plugin) OnConfigurationChange() error {
...
+		Username:    "demoplugin",
+		DisplayName: "Demo Plugin Bot",
+		Description: "A bot account created by the demo plugin.",
+	})
+	if ensureBotError != nil || botId == "" {
```
lieut-data: nit: is it necessary to check `botId`? Presumably the helpers is broken if it returns `"", nil`?

server/configuration_test.go
```diff
@@ -104,11 +104,16 @@ func TestOnConfigurationChange(t *testing.T) {
 			api := test.SetupAPI()
 			defer api.AssertExpectations(t)
 
+			helpers := &plugintest.Helpers{}
```
lieut-data: Work testing a failure to ensure the bot?

server/configuration.go
```diff
@@ -156,7 +156,7 @@ func (p *Plugin) diffConfiguration(newConfiguration *configuration) {
 		}
 
 		if _, err := p.API.CreatePost(&model.Post{
-			UserId:    newConfiguration.demoUserId,
```
lieut-data: Is there an opportunity to remove some of the demo user configuration?

hanzei: Or we could keep the demo user configuration but patch the bot, when it changes.

server/configuration.go
crspeller: Yup. Will fix.

server/configuration_test.go
crspeller: Yes. I should do that.

server/configuration.go
crspeller: It was the only example of using a user there so I left it. Maybe a separate ticket to show a slightly different use case?

server/configuration.go
crspeller: This is a really hard question. I kinda want to say that we can use `error` because it's an entirely new interface, helpers. But you make a good point. Going to make the other changes and think about this.

hanzei: Maybe we can go with the convention of using `err` for `error`'s and `appErr` for `*model.AppError`? This makes it easier to distinguish the two types.

crspeller:
@lieut-data Apologizes for the delay. I have pushed feedback fixes.

PR #30, opened 2019-05-12
MM-14384 ephemeral post function demo

scottleedavis:
Fixes https://github.com/mattermost/mattermost-server/issues/10392

* ephemeral post function demo.  adds `/demo_plugin ephemeral`
* Update to 5.10 mattermost-server dependency


server/command_hooks.go
```diff
@@ -17,10 +17,10 @@ func (p *Plugin) registerCommand(teamId string) error {
 		TeamId:           teamId,
 		Trigger:          CommandTrigger,
 		AutoComplete:     true,
-		AutoCompleteHint: "(true|false)",
-		AutoCompleteDesc: "Enables or disables the demo plugin hooks.",
+		AutoCompleteHint: "(true|false|ephemeral)",
+		AutoCompleteDesc: "Enables or disables the demo plugin hooks, or demonstrates an ephemeral post capabilities.",
 		DisplayName:      "Demo Plugin Command",
-		Description:      "A command used to enable or disable the demo plugin hooks.",
+		Description:      "A command used to enable or disable the demo plugin hooks, or demonstrate ephemeral post capabilities.",
```
hanzei: I would like to add this as a new command. 3/5
Maybe just `/ephemeral`? 1/5

server/http_hooks.go
```diff
@@ -45,3 +51,66 @@ func (p *Plugin) handleStatus(w http.ResponseWriter, r *http.Request) {
...
+		Message:   fmt.Sprintf("updated ephemeral action %d", count),
+		Props: model.StringInterface{
+			"attachments": []*model.SlackAttachment{
+				{
+					Actions: []*model.PostAction{
+						{
+							Id: model.NewId(),
+							Integration: &model.PostActionIntegration{
+								Context: model.StringInterface{
+									"count": count,
```
hanzei: :100:

server/http_hooks.go
```diff
@@ -45,3 +51,66 @@ func (p *Plugin) handleStatus(w http.ResponseWriter, r *http.Request) {
...
 }
+
+func (p *Plugin) handleEphemeralUpdate(w http.ResponseWriter, r *http.Request) {
+	request := model.PostActionIntegrationRequestFromJson(r.Body)
+	URL := fmt.Sprintf("%s", *p.API.GetConfig().ServiceSettings.SiteURL)
+
+	count := int(request.Context["count"].(float64)) + 1
+
+	post := &model.Post{
+		Id:        request.PostId,
```
hanzei: I don't think this is needed. I would rather not be explicit here, as most integrations arn't either.

server/http_hooks.go
```diff
@@ -45,3 +51,66 @@ func (p *Plugin) handleStatus(w http.ResponseWriter, r *http.Request) {
...
+		ChannelId: request.ChannelId,
+		UserId:    request.UserId,
```
hanzei: Same here and with `ChannelId`.

server/http_hooks.go
```diff
@@ -45,3 +51,66 @@ func (p *Plugin) handleStatus(w http.ResponseWriter, r *http.Request) {
...
+						{
+							Id: model.NewId(),
+							Integration: &model.PostActionIntegration{
+								Context: model.StringInterface{
+									"count": count,
+								},
+								URL: fmt.Sprintf("%s/plugins/%s/ephemeral/update", URL, manifest.Id),
+							},
+							Type: model.POST_ACTION_TYPE_BUTTON,
+							Name: "Update",
```
hanzei: Maybe show `count` in this? 0/5

server/command_hooks.go
```diff
@@ -82,6 +82,45 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
...
+		post := &model.Post{
+			ChannelId: args.ChannelId,
+			UserId:    args.UserId,
+			Message:   "test ephemeral actions",
+			Props: model.StringInterface{
+				"attachments": []*model.SlackAttachment{
+					{
+						Actions: []*model.PostAction{
+							{
+								Id: model.NewId(),
```
hanzei: Is this needed? I don't think so.

server/http_hooks.go
```diff
@@ -45,3 +51,66 @@ func (p *Plugin) handleStatus(w http.ResponseWriter, r *http.Request) {
...
+			},
+		},
+	}
+	p.API.UpdateEphemeralPost(request.UserId, post)
+
+	resp := &model.PostActionIntegrationResponse{}
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	_, _ = w.Write(resp.ToJson())
+
```
hanzei: Nit: Remove this line.

server/http_hooks.go
```diff
@@ -45,3 +51,66 @@ func (p *Plugin) handleStatus(w http.ResponseWriter, r *http.Request) {
...
+}
+
+func (p *Plugin) handleEphemeralDelete(w http.ResponseWriter, r *http.Request) {
+	request := model.PostActionIntegrationRequestFromJson(r.Body)
+
+	post := &model.Post{
+		Id: request.PostId,
+	}
+	p.API.DeleteEphemeralPost(request.UserId, post)
```
hanzei: This is 100% correct. But it looks awful. We have to fix this API method some day...

go.sum
```diff
@@ -44,6 +46,8 @@ github.com/magiconair/properties v1.8.0 h1:LLgXmsheXeRoUOBOjtwPQCWIYqM/LU1ayDtDe
 github.com/magiconair/properties v1.8.0/go.mod h1:PppfXfuXeibc/6YijjN8zIbojt8czPbwD3XqdrwzmxQ=
 github.com/mattermost/mattermost-server v5.9.0+incompatible h1:bUlZKAhjp7cGQRepmmsv225vbXyZaXOuVstqgpB+cE4=
 github.com/mattermost/mattermost-server v5.9.0+incompatible/go.mod h1:5L6MjAec+XXQwMIt791Ganu45GKsSiM+I0tLR9wUj8Y=
```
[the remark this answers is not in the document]
scottleedavis: Done.  :)

server/command_hooks.go
scottleedavis: Created a new command, and felt `demo_ephemeral` was a better candidate.  Open to changes.

server/http_hooks.go
scottleedavis: :)

server/http_hooks.go
scottleedavis: Id is needed so UpdateEphemeralPost knows which post to update.  (unless you meant count, which is used to show how many times the post was updated) .  Could you clarify what you meant please?

server/http_hooks.go
scottleedavis: Without `ChannelId` , UpdateEphemeralPost doesn't appear to work for my tests.

server/http_hooks.go
scottleedavis: I really like this idea.  Changed to put count in the update button.

server/command_hooks.go
scottleedavis: Not needed.  Removed Id from each `PostAction`

server/http_hooks.go
scottleedavis: the header?  removed on both that and in handleEphemeralDelete.

server/http_hooks.go
scottleedavis: 100% agreed.  I noticed you brought it up in ~extensions earlier today, so I look forward to the new version.

server/http_hooks.go
scottleedavis: Ah, I think you mean't count.  I updated it to this:
`count := request.Context["count"].(float64) + 1`

server/http_hooks.go
levb: That's true, I ran into the same thing. Worth a ticket, or by design?

scottleedavis: As ChannelId is usually readily available, I am 0/5 if it stays in.  That being said, it seems PostId may be enough by itself?

server/command_hooks.go
hanzei: In #32 `/crash` is used and it quite like this. I'm 3/5 on `/ephemeral` now. It's a short and clear name for what the plugin does. Why do you want to go with `demo_ephemeral`?

server/http_hooks.go
hanzei: After thinking a bit more about this, I would leave it as it is. The only require informations are `Id` and `ChannelId`, correct? That seams fine to me.

server/http_hooks.go
```diff
@@ -45,3 +51,62 @@ func (p *Plugin) handleStatus(w http.ResponseWriter, r *http.Request) {
...
+									"count": count,
+								},
+								URL: fmt.Sprintf("%s/plugins/%s/ephemeral/update", URL, manifest.Id),
+							},
+							Type: model.POST_ACTION_TYPE_BUTTON,
+							Name: fmt.Sprintf("Update %d", int(count)),
+						},
+						{
+							Integration: &model.PostActionIntegration{
+								Context: model.StringInterface{},
```
hanzei: This isn't needed. Mind deleting it?

server/command_hooks.go
```diff
@@ -44,44 +56,77 @@ func (p *Plugin) emitStatusChange() {
...
+										"count": 0,
+									},
+									URL: fmt.Sprintf("%s/plugins/%s/ephemeral/update", URL, manifest.Id),
+								},
+								Type: model.POST_ACTION_TYPE_BUTTON,
+								Name: "Update",
+							},
+							{
+								Integration: &model.PostActionIntegration{
+									Context: model.StringInterface{},
```
hanzei: This isn't needed. Mind deleting it?

server/command_hooks.go
```diff
@@ -44,44 +56,77 @@ func (p *Plugin) emitStatusChange() {
...
+			ChannelId: args.ChannelId,
+			UserId:    args.UserId,
+			Message:   "test ephemeral actions",
+			Props: model.StringInterface{
+				"attachments": []*model.SlackAttachment{
+					{
+						Actions: []*model.PostAction{
+							{
+								Integration: &model.PostActionIntegration{
+									Context: model.StringInterface{
```
hanzei: I would like (2/5) go get rid of `model.StringInterface` in this code base. What do you think about using `map[string]interface{}` instead?

server/http_hooks.go
```diff
@@ -45,3 +51,62 @@ func (p *Plugin) handleStatus(w http.ResponseWriter, r *http.Request) {
...
+					},
+				},
+			},
+		},
+	}
+	p.API.UpdateEphemeralPost(request.UserId, post)
+
+	resp := &model.PostActionIntegrationResponse{}
+	w.WriteHeader(http.StatusOK)
+	_, _ = w.Write(resp.ToJson())
```
hanzei: Is this need? Isn't an empty response enough?

server/http_hooks.go
```diff
@@ -45,3 +51,62 @@ func (p *Plugin) handleStatus(w http.ResponseWriter, r *http.Request) {
...
+
+}
+
+func (p *Plugin) handleEphemeralDelete(w http.ResponseWriter, r *http.Request) {
+	request := model.PostActionIntegrationRequestFromJson(r.Body)
```
hanzei: See above

server/http_hooks.go
hanzei: Well, this got fixed faster then I thought. What do you think about adapting the new API and requiring MM `v5.12.0` for this plugin.

server/command_hooks.go
scottleedavis: I had originally wanted to preserve the 'demo_*' namespace, but if we are going to add `/crash` it makes sense to use `/ephemeral` too.

server/http_hooks.go
scottleedavis: Sounds good.  I tried
```	github.com/mattermost/mattermost-server v5.12.0+incompatible``` in go.mod, but there is an error  `unknown revision v5.12.0`.  Leaving as is, and changing `const minimumServerVersion = "5.12.0"`

Comment (scottleedavis) [reply server/http_hooks.go]: Seems that using minimumServerVersion at 5.12 isn't available yet on master & release-5.12 branches haven't updated version yet.  

Also, when i update go.mod/go.sum to the latest mattermost master, I get these seemingly unrelated errors
```
# github.com/mattermost/mattermost-plugin-demo/server [github.com/mattermost/mattermost-plugin-demo/server.test]
server/configuration.go:189:28: p.Helpers undefined (type *Plugin has no field or method Helpers)
server/configuration_test.go:71:28: undefined: plugintest.Helpers
server/configuration_test.go:86:26: undefined: plugintest.Helpers
server/configuration_test.go:87:17: undefined: plugintest.Helpers
server/configuration_test.go:106:26: undefined: plugintest.Helpers
server/configuration_test.go:107:17: undefined: plugintest.Helpers
server/configuration_test.go:124:26: undefined: plugintest.Helpers
server/configuration_test.go:125:17: undefined: plugintest.Helpers
server/configuration_test.go:145:5: p.SetHelpers undefined (type Plugin has no field or method SetHelpers)
```

What would you like me to do @hanzei ?  Update and fix the changes, or leave this to be fixed in a future PR?  (or other) . happy to do whichever.  :)

(paused changing to the new DeleteEphemeralPost until this is resolved.)

File: server/http_hooks.go [hunk 15]
```diff
@@ -45,3 +51,62 @@ func (p *Plugin) handleStatus(w http.ResponseWriter, r *http.Request) {
...
+func (p *Plugin) handleEphemeralUpdate(w http.ResponseWriter, r *http.Request) {
+	request := model.PostActionIntegrationRequestFromJson(r.Body)
+	URL := fmt.Sprintf("%s", *p.API.GetConfig().ServiceSettings.SiteURL)
+
+	count := request.Context["count"].(float64) + 1
+
+	post := &model.Post{
+		Id:        request.PostId,
+		ChannelId: request.ChannelId,
+		UserId:    request.UserId,
```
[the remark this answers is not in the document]
scottleedavis: Appears not.   Removing.

server/http_hooks.go
```diff
@@ -45,3 +51,62 @@ func (p *Plugin) handleStatus(w http.ResponseWriter, r *http.Request) {
 func (p *Plugin) handleHello(w http.ResponseWriter, r *http.Request) {
 	w.Write([]byte("Hello World!"))
 }
+
+func (p *Plugin) handleEphemeralUpdate(w http.ResponseWriter, r *http.Request) {
+	request := model.PostActionIntegrationRequestFromJson(r.Body)
+	URL := fmt.Sprintf("%s", *p.API.GetConfig().ServiceSettings.SiteURL)
```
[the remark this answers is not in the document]
scottleedavis: Habit I suppose.  using your recommendation.

server/http_hooks.go
```diff
@@ -45,3 +51,62 @@ func (p *Plugin) handleStatus(w http.ResponseWriter, r *http.Request) {
 func (p *Plugin) handleHello(w http.ResponseWriter, r *http.Request) {
 	w.Write([]byte("Hello World!"))
 }
+
+func (p *Plugin) handleEphemeralUpdate(w http.ResponseWriter, r *http.Request) {
+	request := model.PostActionIntegrationRequestFromJson(r.Body)
```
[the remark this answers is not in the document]
scottleedavis: Done

server/command_hooks.go
scottleedavis: Done

server/command_hooks.go
scottleedavis: Using model.StringInterface looks cleaner to me personally.  I can make the switch to `map[string]interface{}` if you'd prefer though.

server/http_hooks.go
scottleedavis: Done

server/http_hooks.go
scottleedavis: Without this response errors appear in the log `Action integration error                      caller="mlog/log.go:172" err_details="err=EOF" err_where=DoPostAction http_code=400 ip_addr="::1" method=POST path=/api/v4/posts/5qz78hib7idpzdo9fun5i4jypo/actions/74377u6mfbbk9kr38orwu1319w request_id=njzege7kcjrqicw7mfbnkzende user_id=mywge9a3ibyd9nosbkaaubo4ph`

server/command_hooks.go
hanzei: I'm only 2/5. Let's go with `model.StringInterface`, if you prefer it.

server/http_hooks.go
hanzei: Didn't know that. Thanks!

server/configuration.go
```diff
@@ -220,7 +220,7 @@ func (p *Plugin) ensureDemoUser(configuration *configuration) (string, error) {
 	if user == nil {
 		user, err = p.API.CreateUser(&model.User{
 			Username:  configuration.Username,
-			Password:  "password",
+			Password:  "P@ssw0rd!!",
```
hanzei: Why is this change needed? :question:

server/http_hooks.go
hanzei: You can update the dependency with `go get github.com/mattermost/mattermost-server@2d3e41783398dc16689e5c1732f9d21727383e2c`. Would you please try this?


We can bump `minimumServerVersion` once DB bump is done. This should be done in the next couple of days.

scottleedavis: That works perfectly @hanzei 👍

server/configuration.go
scottleedavis: Whoops.  I was testing and on a branch there was a 10 character + special characters limit.  Changing this back

server/README.md
```diff
@@ -59,6 +59,9 @@ This demo implementation logs a message to the demo channel whenever a user leav
 This demo implementation responds to a `/demo_plugin` command, allowing the user to enable
 or disable the demo plugin's hooks functionality (but leave the command and webapp enabled).
 
+A third option 'ephemeral' demonstrates ephemeral interactive usage of SendEphemeralPost,
```
hanzei: This has to be updated. Sorry, forgot it in the last review.

scottleedavis: Done.  👍

hanzei:
Thanks again Scott for this contribution :+1:

PR #32, opened 2019-05-21
Mm 15584 - add /crash command that crashes the plugin

scottleedavis:
Fixes https://github.com/mattermost/mattermost-server/issues/10901

* adds /crash command that crashes the plugin by divide by zero error

### DIFF
--- a/server/command_hooks.go
+++ b/server/command_hooks.go
@@ -57,44 +57,45 @@ func (p *Plugin) emitStatusChange() {
 func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*model.CommandResponse, *model.AppError) {
 	configuration := p.getConfiguration()
 
-	if !strings.HasPrefix(args.Command, "/"+CommandTrigger) {
-		return &model.CommandResponse{
-			ResponseType: model.COMMAND_RESPONSE_TYPE_EPHEMERAL,
-			Text:         fmt.Sprintf("Unknown command: " + args.Command),
-		}, nil
-	}
+	if strings.HasPrefix(args.Command, "/"+CommandTrigger) {
 
-	if strings.HasSuffix(args.Command, "true") {
-		if !configuration.disabled {
+		if strings.HasSuffix(args.Command, "true") {
+			if !configuration.disabled {
+				return &model.CommandResponse{
+					ResponseType: model.COMMAND_RESPONSE_TYPE_EPHEMERAL,
+					Text:         "The demo plugin hooks are already enabled.",
+				}, nil
+			}
+
+			configuration.disabled = false
+			p.emitStatusChange()
+
 			return &model.CommandResponse{
 				ResponseType: model.COMMAND_RESPONSE_TYPE_EPHEMERAL,
-				Text:         "The demo plugin hooks are already enabled.",
+				Text:         "Enabled demo plugin hooks.",
 			}, nil
-		}
 
-		configuration.disabled = false
-		p.emitStatusChange()
+		} else if strings.HasSuffix(args.Command, "false") {
+			if configuration.disabled {
+				return &model.CommandResponse{
+					ResponseType: model.COMMAND_RESPONSE_TYPE_EPHEMERAL,
+					Text:         "The demo plugin hooks are already disabled.",
+				}, nil
+			}
 
-		return &model.CommandResponse{
-			ResponseType: model.COMMAND_RESPONSE_TYPE_EPHEMERAL,
-			Text:         "Enabled demo plugin hooks.",
-		}, nil
+			configuration.disabled = true
+			p.emitStatusChange()
 
-	} else if strings.HasSuffix(args.Command, "false") {
-		if configuration.disabled {
 			return &model.CommandResponse{
 				ResponseType: model.COMMAND_RESPONSE_TYPE_EPHEMERAL,
-				Text:         "The demo plugin hooks are already disabled.",
+				Text:         "Disabled demo plugin hooks.",
 			}, nil
 		}
-
-		configuration.disabled = true
-		p.emitStatusChange()
-
-		return &model.CommandResponse{
-			ResponseType: model.COMMAND_RESPONSE_TYPE_EPHEMERAL,
-			Text:         "Disabled demo plugin hooks.",
-		}, nil
+	} else if strings.HasPrefix(args.Command, "/"+CommandTriggerCrash) {
+		y := 0
+		x := 5 / y
+		p.API.LogError("{}", x)
+		return &model.CommandResponse{}, nil
 	}
 
 	return &model.CommandResponse{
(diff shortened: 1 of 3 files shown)


scottleedavis:
I am confused as to the build failure.  The go.mod and go.sum are identical to the master branch.  The build `make` runs without error locally.

jasonblais:
Looks like the build failure got resolved 👍 

@hanzei please help assign appropriate reviewers

hanzei:
@scottleedavis I rerun the build and this fixed it. Don't know why it failed in the first place.

server/command_hooks.go
```diff
@@ -25,6 +26,18 @@ func (p *Plugin) registerCommand(teamId string) error {
 		return errors.Wrap(err, "failed to register command")
 	}
 
+	if err := p.API.RegisterCommand(&model.Command{
+		TeamId:           teamId,
+		Trigger:          CommandTriggerCrash,
+		AutoComplete:     true,
+		AutoCompleteHint: "",
+		AutoCompleteDesc: "Crashes plugin.",
+		DisplayName:      "Crashes plugin",
```
hanzei: `DisplayName` and `Description` are not needed. They are only used for "normal" integrations.

server/command_hooks.go
```diff
@@ -25,6 +26,18 @@ func (p *Plugin) registerCommand(teamId string) error {
 		return errors.Wrap(err, "failed to register command")
 	}
 
+	if err := p.API.RegisterCommand(&model.Command{
+		TeamId:           teamId,
+		Trigger:          CommandTriggerCrash,
+		AutoComplete:     true,
+		AutoCompleteHint: "",
+		AutoCompleteDesc: "Crashes plugin.",
```
hanzei:
```suggestion
		AutoCompleteDesc: "Crashes Demo Plugin",
```

server/command_hooks.go
```diff
@@ -44,44 +57,45 @@ func (p *Plugin) emitStatusChange() {
...
-			Text:         "Disabled demo plugin hooks.",
-		}, nil
+	} else if strings.HasPrefix(args.Command, "/"+CommandTriggerCrash) {
+		y := 0
+		x := 5 / y
+		p.API.LogError("{}", x)
+		return &model.CommandResponse{}, nil
 	}
 
 	return &model.CommandResponse{
```
hanzei: This should be the block from above:

```suggestion
	return &model.CommandResponse{
		ResponseType: model.COMMAND_RESPONSE_TYPE_EPHEMERAL,
		Text:         fmt.Sprintf("Unknown command: " + args.Command),
	}, nil
```

server/command_hooks.go
```diff
@@ -44,44 +57,45 @@ func (p *Plugin) emitStatusChange() {
...
+			p.emitStatusChange()
 
-	} else if strings.HasSuffix(args.Command, "false") {
-		if configuration.disabled {
 			return &model.CommandResponse{
 				ResponseType: model.COMMAND_RESPONSE_TYPE_EPHEMERAL,
-				Text:         "The demo plugin hooks are already disabled.",
+				Text:         "Disabled demo plugin hooks.",
 			}, nil
 		}
```
hanzei: You have to check if the two other cases where missed:

```suggestion
		}
		
		return &model.CommandResponse{
			ResponseType: model.COMMAND_RESPONSE_TYPE_EPHEMERAL,
			Text:         fmt.Sprintf("Unknown command: " + args.Command),
		}, nil
```

server/command_hooks.go
```diff
@@ -44,44 +57,45 @@ func (p *Plugin) emitStatusChange() {
...
 
-	if !strings.HasPrefix(args.Command, "/"+CommandTrigger) {
-		return &model.CommandResponse{
-			ResponseType: model.COMMAND_RESPONSE_TYPE_EPHEMERAL,
-			Text:         fmt.Sprintf("Unknown command: " + args.Command),
-		}, nil
-	}
+	if strings.HasPrefix(args.Command, "/"+CommandTrigger) {
+
+		if strings.HasSuffix(args.Command, "true") {
```
hanzei: Maybe use a `switch`? 1/5

server/command_hooks.go
```diff
@@ -44,44 +57,45 @@ func (p *Plugin) emitStatusChange() {
...
 	configuration := p.getConfiguration()
 
-	if !strings.HasPrefix(args.Command, "/"+CommandTrigger) {
-		return &model.CommandResponse{
-			ResponseType: model.COMMAND_RESPONSE_TYPE_EPHEMERAL,
-			Text:         fmt.Sprintf("Unknown command: " + args.Command),
-		}, nil
-	}
+	if strings.HasPrefix(args.Command, "/"+CommandTrigger) {
+
```
hanzei: Nit: remove this line

server/command_hooks.go
```diff
@@ -44,44 +57,45 @@ func (p *Plugin) emitStatusChange() {
...
-		p.emitStatusChange()
-
-		return &model.CommandResponse{
-			ResponseType: model.COMMAND_RESPONSE_TYPE_EPHEMERAL,
-			Text:         "Disabled demo plugin hooks.",
-		}, nil
+	} else if strings.HasPrefix(args.Command, "/"+CommandTriggerCrash) {
+		y := 0
+		x := 5 / y
+		p.API.LogError("{}", x)
```
hanzei: This doesn't get executed, because the plugin fails first. Why is this there?

server/command_hooks.go
```diff
@@ -44,44 +57,45 @@ func (p *Plugin) emitStatusChange() {
...
-		configuration.disabled = true
-		p.emitStatusChange()
-
-		return &model.CommandResponse{
-			ResponseType: model.COMMAND_RESPONSE_TYPE_EPHEMERAL,
-			Text:         "Disabled demo plugin hooks.",
-		}, nil
+	} else if strings.HasPrefix(args.Command, "/"+CommandTriggerCrash) {
+		y := 0
+		x := 5 / y
```
hanzei: Maybe this should be an own go routine? Currently the UX is a bit weird, because the API doesn't finish.
Screenshot from 2019-05-27 15-47-10

server/command_hooks.go
scottleedavis: Removed.

server/command_hooks.go
scottleedavis: This line is needed though to check between the `/demo_plugin` and `/crash` commands...   Am I misunderstanding what you are asking?

server/command_hooks.go
scottleedavis: Maybe combing this to a switch after `/ephemeral` is merged?  Both plugins use the if..else approach, and likely would be cleaner if the three were in a switch (and further commands)

server/command_hooks.go
scottleedavis: So that the plugin would build.  I changed it to be this 
```
		y := 0
		y = 5 / y
```

server/command_hooks.go
scottleedavis: put it i it's own go routine, and an output 'crashing plugin' for the command response

server/command_hooks.go
hanzei: I was talking about the empty line. But never mind, we can fix this later.

hanzei:
@scottleedavis Would you please merge `master` into this?

hanzei:
@cpoile I know you are busy at the moment. It would be great if you have the time to review this.

scottleedavis:
Done @hanzei  👍

scottleedavis:
thank you @cpoile and @hanzei !  :)

PR #33, opened 2019-05-27
MM-15020 Adding server localization

crspeller:
Implementation of: https://github.com/mattermost/mattermost-server/pull/10949

#### DRAFT: 
- Needs the server implementation

### DIFF
--- a/server/command_hooks.go
+++ b/server/command_hooks.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"fmt"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -47,7 +46,7 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 	if !strings.HasPrefix(args.Command, "/"+CommandTrigger) {
 		return &model.CommandResponse{
 			ResponseType: model.COMMAND_RESPONSE_TYPE_EPHEMERAL,
-			Text:         fmt.Sprintf("Unknown command: " + args.Command),
+			Text:         p.Helpers.TContext(c, "command.unknown_command", map[string]interface{}{"Command": args.Command}),
 		}, nil
 	}
 
@@ -55,7 +54,7 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 		if !configuration.disabled {
 			return &model.CommandResponse{
 				ResponseType: model.COMMAND_RESPONSE_TYPE_EPHEMERAL,
-				Text:         "The demo plugin hooks are already enabled.",
+				Text:         p.Helpers.TContext(c, "command.already_enabled"),
 			}, nil
 		}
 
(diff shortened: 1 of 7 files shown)


server/command_hooks.go
```diff
@@ -80,12 +79,12 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
...
 			ResponseType: model.COMMAND_RESPONSE_TYPE_EPHEMERAL,
-			Text:         "Disabled demo plugin hooks.",
+			Text:         p.Helpers.TContext(c, "command.disabled"),
 		}, nil
 	}
 
 	return &model.CommandResponse{
 		ResponseType: model.COMMAND_RESPONSE_TYPE_EPHEMERAL,
-		Text:         fmt.Sprintf("Unknown command action: " + args.Command),
+		Text:         p.Helpers.TContext(c, "command.unknown_command", map[string]interface{}{"Command": args.Command}),
```
cpoile: I guess there's no easier way to send the args?

server/configuration.go
```diff
@@ -181,6 +181,12 @@ func (p *Plugin) OnConfigurationChange() error {
 		return errors.Wrap(loadConfigErr, "failed to load plugin configuration")
 	}
 
+	// Setup translations
+	config := p.API.GetConfig()
+	if locError := p.Helpers.SetLocalization(&config.LocalizationSettings, "plugins/"+manifest.Id+"/assets"); err != nil {
```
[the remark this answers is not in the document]
hmhealey: Maybe the path could be defined in the plugin manifest?

hmhealey: 0/5, but maybe we should put the translation files in `assets/i18n` to keep them consistent with the other repos?

hanzei: Maybe even `assets/i18n/server`?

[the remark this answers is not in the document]
crspeller: Ya. I tried some stuff, but everything was just over-complicating the situation. It's really just one line to copy from our template. I don't think it's a lot to ask.  

We don't actually have access to the manifest from the plugin helpers where we need it.

server/command_hooks.go
crspeller: It's how it's done on the server. Keeping the same API to be consistent. 
It has been proposed alternatively to forget what the server does and upgrade to v2 of the i18n library we use: https://github.com/nicksnyder/go-i18n
Thoughts?

server/configuration.go
crspeller: Agree. Will change. I don't think `/server` is necessary though because the webapp translations aren't in assets (they get bundled with webpack).

server/command_hooks.go
hanzei: FIY: I commented about this topic in the server PR https://github.com/mattermost/mattermost-server/pull/10949#issuecomment-497027177

jasonblais:
Awaiting https://github.com/mattermost/mattermost-server/pull/10949

jasonblais:
As noted in the server PR, this has to be redone because of the refactor of the server to use the new version of react-intl. Jesse's team will look at cleaning up the ticket and merge with existing epics or efforts via https://mattermost.atlassian.net/browse/MM-15020. Closing.

jasonblais:
Or...I can't close it so @crspeller I'll let you do the honours ;)

cpoile:
Ohh oohh, can I do it? :)

PR #34, opened 2019-05-28
Note that the demo plugin is compatible for v5.10+

jasonblais:
We should also cut a release and clean up the releases page at some point.

PR that increased server compatibility to 5.10+: https://github.com/mattermost/mattermost-plugin-demo/pull/29

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,6 @@ This plugin demonstrates the capabilities of a Mattermost plugin. It includes th
 
 Feel free to base your own plugin off this repository, removing or modifying components as needed. If you're already familiar with what plugins can do, consider starting from [mattermost-plugin-sample](https://github.com/mattermost/mattermost-plugin-sample) instead, which includes the same build framework but omits the demo implementations.
 
-Note that this plugin is authored for Mattermost 5.2 and later, and is not compatible with earlier releases of Mattermost.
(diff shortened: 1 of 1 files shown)


hanzei:
Why don't we just document this in the releases?

jasonblais:
We can, but I don't know which release tag is compatible which version

hanzei:
We documented this for the last release. See https://github.com/mattermost/mattermost-plugin-demo/releases/tag/v0.0.5. In the future we can continue to keep track of this.

jasonblais:
@hanzei Sure - but if someone builds from source, it's not clear 5.10 (or 5.12?) is required because of https://github.com/mattermost/mattermost-plugin-demo/pull/29

hanzei:
>  Sure - but if someone builds from source, it's not clear 5.10 (or 5.12?) is required because of #29

I see. But I would rather document and enforce this via the manifest. See #36 for the PR.

>  I'm okay cutting a new release tag, v0.0.6, which has the new compatibility requirement documented

:+1:

jasonblais:
@hanzei 

> I see. But I would rather document and enforce this via the manifest.

That's fair. Maybe a good note to a best practice document?

hanzei:
> Maybe a good note to a best practice document?

Surely :+1:

PR #36, opened 2019-06-10
Require Mattermost Server v5.12

hanzei:
See #34 for a discussion about this

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -3,7 +3,7 @@
     "name": "Demo Plugin",
     "description": "This plugin demonstrates the capabilities of a Mattermost plugin.",
     "version": "0.0.6",
-    "min_server_version": "5.10.0",
+    "min_server_version": "5.12.0",
     "server": {
         "executables": {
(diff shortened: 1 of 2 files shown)


hanzei:
@cpoile Not really. If you don't use any of the "new" or changes API method, your plugin will run fine on older version.

But speaking of changed API methods: https://github.com/mattermost/mattermost-plugin-demo/pull/30 makes use of the changed of the changes signature of `DeleteEphemeralMessage`, it will require `v5.12`. Will update.

cpoile:
Actually, I was worried about this line in the go.mod:
```
	github.com/mattermost/mattermost-server v0.0.0-20190516103213-2d3e41783398
```
I don't understand why we're building against a hash, and wonder if it will cause problems?

hanzei:
> I don't understand why we're building against a hash, and wonder if it will cause problems?

We need to build against the hash because we want to make use of the new plugin helper. I know, this is a bit weird because we are using the Server as a library and an API endpoint.

cpoile:
Right, nevermind, I get it now. Friday brain.

PR #40, opened 2019-06-23
added ephemeral_override command

scottleedavis:
adds an `/ephemeral_override` command to demonstrate overriding an ephemeral post with the webapp using a `type` prop.

#### Links
https://github.com/mattermost/mattermost-server/issues/10797

### DIFF
--- a/server/command_hooks.go
+++ b/server/command_hooks.go
@@ -12,10 +12,11 @@ import (
 )
 
 const (
-	commandTriggerCrash     = "crash"
-	commandTriggerHooks     = "demo_plugin"
-	commandTriggerDialog    = "dialog"
-	commandTriggerEphemeral = "ephemeral"
+	commandTriggerCrash             = "crash"
+	commandTriggerHooks             = "demo_plugin"
+	commandTriggerDialog            = "dialog"
+	commandTriggerEphemeral         = "ephemeral"
+	commandTriggerEphemeralOverride = "ephemeral_override"
 
 	dialogElementNameNumber = "somenumber"
 	dialogElementNameEmail  = "someemail"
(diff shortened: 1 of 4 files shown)


hanzei:
Hey Scott,

Thanks for the PR. It great to have this example server side. 

I'm not sure what the expected behaviour for the serve side is. Should they show up "special"? I currently see this:
Screenshot from 2019-06-24 10-58-24

scottleedavis:
Hi @hanzei ,

I get the same plain post as you when testing on a 5.12 release, but on master the following should appear.

Screen Shot 2019-06-24 at 8 39 02 AM

hanzei:
Hm, seams like I tested with an outdated version. Works fine :+1:

server/command_hooks.go
```diff
@@ -183,6 +195,17 @@ func (p *Plugin) executeCommandEphemeral(args *model.CommandArgs) *model.Command
 	return &model.CommandResponse{}
 }
 
+func (p *Plugin) executeCommandEphemeralOverride(args *model.CommandArgs) *model.CommandResponse {
+	_ = p.API.SendEphemeralPost(args.UserId, &model.Post{
+		ChannelId: args.ChannelId,
+		Message:   "test ephemeral override",
+		Props: model.StringInterface{
+			"type": "custom_demo_plugin",
```
marianunez: I had the same confusion as @hanzei of what I should be seeing when I ran the command.

It would be best if we can add another type of post that is different from the one used in the regular posts so the example can be more obvious that the ephemeral messages can have their own custom types as well.

scottleedavis: @marianunez  I updated the example to use it's own post_type aside from the existing post_type, though borrowed most of the content.  Does the following look ok?
Screen Shot 2019-06-25 at 8 31 40 PM

marianunez: This looks great @scottleedavis!

PR #42, opened 2019-06-28
Updating dependencies.

crspeller:


go.mod
```diff
@@ -5,22 +5,23 @@ go 1.12
...
 	github.com/go-ldap/ldap v3.0.3+incompatible // indirect
-	github.com/hashicorp/go-hclog v0.9.0 // indirect
+	github.com/hashicorp/go-hclog v0.9.2 // indirect
+	github.com/hashicorp/go-plugin v1.0.1 // indirect
 	github.com/lib/pq v1.1.1 // indirect
-	github.com/mattermost/go-i18n v1.10.0 // indirect
-	github.com/mattermost/mattermost-server v0.0.0-20190516103213-2d3e41783398
-	github.com/nicksnyder/go-i18n v1.10.0 // indirect
+	github.com/mattermost/go-i18n v1.11.0 // indirect
+	github.com/mattermost/mattermost-server v0.0.0-20190624195857-dbe330511eb4
```
hanzei: This can just be `v5.12.0+incompatible`

go.mod
```diff
@@ -5,22 +5,23 @@ go 1.12
...
 	github.com/stretchr/objx v0.2.0 // indirect
 	github.com/stretchr/testify v1.3.0
 	go.uber.org/atomic v1.4.0 // indirect
 	go.uber.org/zap v1.10.0 // indirect
-	golang.org/x/crypto v0.0.0-20190506204251-e1dfcc566284 // indirect
-	golang.org/x/net v0.0.0-20190503192946-f4e77d36d62c // indirect
-	golang.org/x/sys v0.0.0-20190508100423-12bbe5a7a520 // indirect
-	google.golang.org/genproto v0.0.0-20190508193815-b515fa19cec8 // indirect
-	google.golang.org/grpc v1.20.1 // indirect
+	golang.org/x/crypto v0.0.0-20190621222207-cc06ce4a13d4 // indirect
```
hanzei: I'm a bit surprised this indirect dependencies still show up. Should't they be removed?

webapp/package.json
```diff
@@ -11,26 +11,29 @@
...
+    "eslint-plugin-react": "7.14.2",
+    "file-loader": "4.0.0",
+    "webpack": "4.35.0",
+    "webpack-cli": "3.3.5"
   },
   "dependencies": {
-    "mattermost-redux": "^1.2.0",
-    "react": "16.4.1",
-    "react-intl": "2.8.0",
+    "mattermost-redux": "5.12.0",
```
hanzei: :+1:

webapp/package.json
```diff
@@ -11,26 +11,29 @@
...
-    "babel-preset-env": "^1.7.0",
-    "babel-preset-react": "^6.24.1",
-    "eslint": "^5.2.0",
-    "eslint-import-resolver-webpack": "^0.10.1",
-    "eslint-plugin-import": "^2.13.0",
-    "eslint-plugin-react": "^7.10.0",
-    "file-loader": "3.0.1",
-    "webpack": "^4.16.1",
-    "webpack-cli": "^3.0.8"
+    "@babel/cli": "7.4.4",
```
hanzei: This wasn't required previously. Why is it now?

crspeller: I think they split the package out in the upgrade to 7.

go.mod
crspeller: I don't edit these manually I just use `go mod tidy`.

go.mod
crspeller: Any idea how to get that effect with `go get`? Whenever I try `go get -u github.com/mattermost/mattermost-server` it gives me 5.11. Probably something to do with minimum version selection. So this is `go get -u github.com/mattermost/mattermost-server@release-5.12`. 

I did try `go get -u github.com/mattermost/mattermost-server@v5.12` but that just fails.

webapp/package.json
hanzei: :ok:

go.mod
hanzei: I've run into this problem before. (https://github.com/mattermost/mattermost-plugin-autotranslate/pull/13#discussion_r294298918)

`go mod edit -require github.com/mattermost/mattermost-server@v5.12.0+incompatible` works. Yes, the `+incompatible is required.  (But don't ask me why)

go.mod
hanzei: Not blocking: Why aren't you deleting these manually? It _seams_ to me that `go mod tidy` should do this, but doesn't due to a bug.

go.mod
crspeller: I thought we discussed this before :P 
Pushed an update after running your command. Some things about go modules still confuse me.

go.mod
crspeller: I thought the idea of these files is that they are managed by the `go mod` command. I am inclined just to trust whatever `go mod tidy` thinks is right. Kinda like I trust `gofmt` (or really `goimports`).

hanzei: I agree. Let's leave this as it is and see what go1.13 brings.

PR #44, opened 2019-07-03
[MM-11829] Add example of custom admin console for plugins

krjn:
https://mattermost.atlassian.net/browse/MM-11829

Adds a custom console with a "Hello, world" component to a new admin section, "Plugins (extra)".

Related to mattermost/mattermost-server#10320

Depends on mattermost/mattermost-webapp#2794

### DIFF
--- a/webapp/src/plugin.jsx
+++ b/webapp/src/plugin.jsx
@@ -9,6 +9,7 @@ import es from 'i18n/es.json';
 import {id as pluginId} from './manifest';
 
 import Root from './components/root';
+import AdminSettings from './components/admin_settings';
 import BottomTeamSidebar from './components/bottom_team_sidebar';
 import LeftSidebarHeader from './components/left_sidebar_header';
 import LinkTooltip from './components/link_tooltip';
@@ -42,6 +43,38 @@ function getTranslations(locale) {
     return {};
 }
 
+// eslint-disable-next-line no-unused-vars
+function addToAdminConsole(registry) {
+    registry.registerAdminConsolePlugin((adminDefinition) => {
+        // the settings will appear on the authentication section
+        const demoConfig = {
+            url: 'plugins/demo',
+            icon: 'fa-rocket',
+            title: 'com.mattermost.demo-plugin.title',
+            title_default: 'Demo plugin admin',
+            isHidden: () => false,
+            schema: {
+                id: 'DemoPlugin',
+                component: AdminSettings,
+            },
+        };
+
+        // this is not required - demoConfig could be embedded elsewhere
+        adminDefinition.pluginsExtra = {
+            icon: 'fa fa-plug',
+            id: 'plugins-extra',
+            sectionTitle: 'com.mattermost.demo-plugin.sectionTitle',
+            sectionTitleDefault: 'Plugins (extra)',
+            demo: demoConfig,
+        };
+
+        // or
+        // adminDefinition.authentication.demo = demoConfig;
+
+        return adminDefinition;
+    });
+}
+
 export default class DemoPlugin {
     initialize(registry, store) {
         registry.registerRootComponent(Root);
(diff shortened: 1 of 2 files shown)


webapp/src/plugin.jsx
```diff
@@ -42,6 +43,38 @@ function getTranslations(locale) {
...
+            title_default: 'Demo plugin admin',
+            isHidden: () => false,
+            schema: {
+                id: 'DemoPlugin',
+                component: AdminSettings,
+            },
+        };
+
+        // this is not required - demoConfig could be embedded elsewhere
+        adminDefinition.pluginsExtra = {
```
jespino: I'm not sure if this is the approach that we want to promote. maybe we prefer something like:

```js
adminDefinition.plugins['demo'] = demoConfig
```

krjn: If I'm recalling correctly, currently the plugins section is special, and so that wouldn't work. I agree though, and believe we should perhaps confine the plugins admin consoles to this section after it accepts custom consoles. Additionally, those with custom sections should probably have the default section (from the schema/config) inhibited.

https://github.com/mattermost/mattermost-webapp/blob/master/components/admin_console/admin_sidebar/admin_sidebar.jsx#L204

jespino: The idea of have this "full access" is because we want super powerful plugins, if you don't want to use a plugin, you can just don't use it, but if you use a plugin, you must know that it can do a lot of things. In this case, edit entirely the admin console. For example, I would like to create a plugin that hides sections, so... you can simplify the admin console removing enything that is "distracting for you".

About the plugin sections special case, the plugins section will always render the plugins config items, but you can still add things to it, because it will show the items in the admin definitions too.

lieut-data:
@krjn, what are your thoughts on @jespino's comments above? I've had a chance to review the corresponding webapp pull request, and now understand the motivation here better.

Ideally, for the demo plugin, we would just modify the reducer to add a new custom component.

krjn:
@lieut-data I'm not clear on what is in question. If it's whether we should add a new section in the demo or not, then what I think I remember is that we can't add a normal entry outside of a section, as the implementation that generates the menu does not take that into account - hence the commented code that adds the demo to the authentication section.

jespino:
@lieut-data is this ready to be merged, apart from the merge conflict?

lieut-data:
@jespino, I don't think this quite "illustrates" anything for a plugin developer yet, since it's not obvious how to use it to customize the system console for the plugin in question. I'm inclined to wait for https://mattermost.atlassian.net/browse/MM-18089 and update this to illustrate same. Thoughts?

jespino:
@lieut-data good point, because it is a lower level API, that would be a better option to illustrate more common use cases for developers :)

hanzei:
@krjn Would you please resolve the merge conflict?

lieut-data:
@jespino / @hanzei, wondering if we should defer this change for now and rely on @marianunez's illustration of same in the demo plugin.

hanzei:
To me the demo plugin should present the different ways a plugin can use tool provided by Mattermost. https://github.com/mattermost/mattermost-webapp/pull/2794 added such tool and hence it should be shown in this plugin. 1/5

lieut-data:
@hanzei, in general, I fully agree, except that we haven't yet found a "meaningful" implementation of this feature that isn't just a custom component we're promoting using other means. Perhaps if we could find a different way to illustrate?

hanzei:
Okay, I see. Would a more "meaningful" demonstration be e.g. a fully custom page?

lieut-data:
That's definitely one approach. @jespino, do you have an idea for what we should demo using this low-level feature?

jasonblais:
@jespino Quick nudge on the above question from @lieut-data 

> @jespino, do you have an idea for what we should demo using this low-level feature?

jespino:
@lieut-data We can do things like prefix all sections names in the admin with the "[DemoPlugin] " text, or uppercase everything in the menu, or even simpler, add a new menu section and entry, but using the low level api.

lieut-data:
@jespino, I'm inclined to close this PR out for now until there's demand for the really low-level functionality. Any thoughts?

jespino:
@lieut-data sounds good to me

PR #49, opened 2019-08-01
MM-10982 Extend "/dialog" command to generate interactive dialog with relative callback URL

saturninoabril:
Not mention in the ticket but it might be helpful to extend `/dialog` command to have relative callback URL when trying interactive dialog.

Note that this demo plugin may be used for end-to-end testing in the future.

Relative URL as seen in request payload:
Screen Shot 2019-08-01 at 11 29 01 PM

Command response once submitted:
Screen Shot 2019-08-01 at 11 28 52 PM

### DIFF
--- a/server/command_hooks.go
+++ b/server/command_hooks.go
@@ -24,6 +24,7 @@ const (
 	commandDialogHelp = "###### Interactive Dialog Slash Command Help\n" +
 		"- `/dialog` - pen an Interactive Dialog. Once submitted, user-entered input is posted back into a channel.\n" +
 		"- `/dialog no-elements` - Open an Interactive Dialog with no elements. Once submitted, user's action is posted back into a channel.\n" +
+		"- `/dialog relative-callback-url` - Open an Interactive Dialog with relative allback URL. Once submitted, user's action is posted back into a channel.\n" +
 		"- `/dialog help` - Show this help text"
 )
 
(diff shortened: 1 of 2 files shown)


server/command_hooks.go
```diff
@@ -24,6 +24,7 @@ const (
 	commandDialogHelp = "###### Interactive Dialog Slash Command Help\n" +
 		"- `/dialog` - pen an Interactive Dialog. Once submitted, user-entered input is posted back into a channel.\n" +
 		"- `/dialog no-elements` - Open an Interactive Dialog with no elements. Once submitted, user's action is posted back into a channel.\n" +
+		"- `/dialog relative-callback-url` - Open an Interactive Dialog with relative allback URL. Once submitted, user's action is posted back into a channel.\n" +
```
lieut-data:
```suggestion
		"- `/dialog relative-callback-url` - Open an Interactive Dialog with relative callback URL. Once submitted, user's action is posted back into a channel.\n" +
```

server/http_hooks.go
```diff
@@ -144,12 +144,17 @@ func (p *Plugin) handleDialog2(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	suffix := ""
+	if request.State == "relative-callback-url" {
```
hanzei: Would you please make this a constant? It's used in three locations.

saturninoabril: Sure, I added dialog state constants.  Please take a look again.

PR #51, opened 2019-08-13
MM-17438: attach file to post

lieut-data:
Plugins weren't able to successfully attach a file to a post in 5.12 because a unintentional security change. Verify that this is now working by attaching the demo plugin's configuration as a `.json` attachment to the `OnConfigChange: loading new configuration` posts it makes in the demo plugin channel.

### DIFF
--- a/server/configuration.go
+++ b/server/configuration.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


server/configuration.go
```diff
@@ -155,12 +156,25 @@ func (p *Plugin) diffConfiguration(newConfiguration *configuration) {
 			continue
 		}
 
+		newConfigurationData, jsonErr := json.Marshal(newConfiguration)
+		if jsonErr != nil {
+			p.API.LogWarn("failed to marshal new configuration", "err", err)
+			return
+		}
+
+		fileInfo, err := p.API.UploadFile(newConfigurationData, demoChannelId, "configuration.json")
```
ali-farooq0: are files tied to a user, and only that user is allowed to attach a file with a post or are files tied to a channel and anybody in that channel can tie that file to a post?

lieut-data: Files may be attached to a user (via the `CreatorId`) or to `nouser`. They are always bound to a channel (not 100% certain why!). Files bound to a specific user may only be used by that user. Files bound to `nouser` may be used by anyone. However, once a file is bound to a post, it cannot be rebound.

PR #52, opened 2019-08-13
MM-15340: dialog without trigger id

lieut-data:
Illustrate how to invoke an interactive dialog without a trigger id, e.g. from a main menu option vs. a typical slash command interaction.

See https://mattermost.atlassian.net/browse/MM-15340


jfrerich:
@lieut-data CI is failing, otherwise, lgtm

PR #53, opened 2019-08-13
MM-14273: demo boolean selector

lieut-data:
Illustrate a boolean selector in an interactive dialog:
image

image


hanzei:
@lieut-data Forgot to mention this in the review: Would you be open on updating https://github.com/mattermost/mattermost-plugin-demo/blob/master/server/docs/dialogs.md#dialog with a new image?

lieut-data:
@hanzei, done: https://github.com/mattermost/mattermost-plugin-demo/pull/56

PR #59, opened 2019-08-21
MM-16903 - Override LHS bot icon in demo plugin

QamarFarooq:
- added an svg bot icon image
- setting bot icon image on config change event
- fixes https://github.com/mattermost/mattermost-server/issues/11731


server/configuration.go
```diff
@@ -209,8 +210,19 @@ func (p *Plugin) OnConfigurationChange() error {
 		return errors.Wrap(ensureBotError, "failed to ensure demo bot.")
 	}
 
+	config := p.API.GetConfig()
+	data, err := ioutil.ReadFile(*config.PluginSettings.Directory + "/com.mattermost.demo-plugin/assets/github.svg")
```
ali-farooq0: I think it'd be better to use the `pluginID` as opposed to hardcoding it.

server/configuration.go
```diff
@@ -209,8 +210,19 @@ func (p *Plugin) OnConfigurationChange() error {
...
+	if err != nil {
+		return errors.Wrap(err, "failed to read bot image")
+	}
+
 	p.botId = botId
 
+	AppError := p.API.SetBotIconImage(botId, data)
```
ali-farooq0: This will be called every time there is a config, lets check `p.API.GetBotIconImage` to see if the bot already has an icon set and then set if if it doesn't. It's kinda awkward doing that but will be addressed with https://mattermost.atlassian.net/browse/MM-17297

server/configuration.go
```diff
@@ -209,8 +210,19 @@ func (p *Plugin) OnConfigurationChange() error {
...
+	if AppError != nil {
```
ali-farooq0: golang convention is to use camelCase. `appErr`

server/configuration.go
ali-farooq0: Also looks like there is an api for you to get the path `p.API.GetBundlePath`.

hanzei: You can just use https://developers.mattermost.com/extend/plugins/server/reference/#API.GetBundlePath instead.

server/configuration.go
```diff
@@ -209,8 +211,32 @@ func (p *Plugin) OnConfigurationChange() error {
...
+	if appErr != nil {
+		if appErr.StatusCode != http.StatusNotFound {
+			return errors.Wrap(appErr, "failed to get bot icon image")
+		}
+
+		appErr = p.API.SetBotIconImage(botId, data)
+		if appErr != nil {
+			return errors.Wrap(appErr, "failed to set bot icon")
+		}
+
```
ali-farooq0: nit: please remove extra line.

QamarFarooq:
> api
you mean to check if the image is already set before setting it? or do you mean for me to add a test case which deals with the case where image is already set


> @QamarFarooq looks great! Can you also please add a unit test to cover the case where image is already set? i.e. when `api.On("GetBotIconImage", mock.Anything).Return([]byte{}, nil)` happens

ali-farooq0:
> 
> 
> do you mean for me to check if the image is already set before setting it? or do you mean to add a test case which deals with the case where image is already set
> 
> > @QamarFarooq looks great! Can you also please add a unit test to cover the case where image is already set? i.e. when `api.On("GetBotIconImage", mock.Anything).Return([]byte{}, nil)` happens

@QamarFarooq to add a test that covers the case where the image is already set.

PR #62, opened 2019-08-26
MM-16946 SubMenu in post dropdown menu 

scottleedavis:
###### Overview

```javascript
        const rootRegisterMenuItem = registry.registerPostDropdownMenuItem(
            <FormattedMessage
                id='submenu.menu'
                defaultMessage='Submenu Example'
            />
        );

        const firstItem = (
            <FormattedMessage
                id='submenu.first'
                key='submenu.first'
                defaultMessage='First Item'
            />
        );
        const sub1RegisterMenuItem = rootRegisterMenuItem(
            firstItem,
            () => {
                store.dispatch(postDropdownSubMenuAction(firstItem));
            }
        );
```

Adds an example submenu using registerPostDropdownMenuItem.

img

###### Related PR
* https://github.com/mattermost/mattermost-webapp/pull/3357


hanzei:
No problem Scott. They will get squashed on merge.

scottleedavis:
Made an update that fixes a bug @saturninoabril found in the PR for webapp.  Perhaps it solves the issue for your too @prapti ?

prapti:
Cool. I'll give it a try :) Thanks for the update @scottleedavis!

webapp/i18n/es.json
```diff
@@ -10,5 +10,9 @@
   "sidebar.enabled": "Habilitado",
   "tooltip.message": "Esta es una información sobre herramientas personalizada del complemento de demostración",
   "useractions.action": "Acción",
-  "useractions.demo": "Complemento de Demostración: "
+  "useractions.demo": "Complemento de Demostración: ",
+  "submenu.menu": "Ejemplo de submenú",
+  "submenu.first": "Primer iteam",
```
jespino: `item` not `iteam` but probably it should sound more natural in spanish like `elemento`.

webapp/src/components/root/root.jsx
```diff
@@ -2,11 +2,32 @@ import React from 'react';
...
-const Root = ({visible, close, theme}) => {
+import {menu} from '../../sub_menu';
+
+const Root = ({visible, close, theme, subMenuId}) => {
     if (!visible) {
         return null;
     }
 
+    let extraContent = '';
+    if (subMenuId) {
```
jespino: I think you are keeping this content after close, so if you open the root modal from any other place, you are seeing the "submenu" extra content here.

By the way, probably the extra content should be more descriptive about himself. You can see there "Third item" without any context, probably you should have something like `Element clicked in the menu: Third item`

webapp/i18n/es.json
scottleedavis: Done

webapp/src/components/root/root.jsx
scottleedavis: Great catch, I solved it by dispatching an empty subMenuId on normal Root modal trigger.  Also added the additional descriptive text.

scottleedavis:
@jespino I simplified the data handling of the submenu structure in the root modal https://github.com/mattermost/mattermost-plugin-demo/pull/62/files#diff-d671a5593a8e810fd52c9f9f0361a48aR21-R25 Does this look ok to you?

prapti:
Leaving the QA Review label on for now until all devs approve to make sure that this is tested again if there's any significant changes.

jespino:
@scottleedavis yes, I think that addresses my concerns. Thanks! :)

webapp/src/actions.js
```diff
@@ -4,6 +4,20 @@ import {id as pluginId} from './manifest';
 import {STATUS_CHANGE, OPEN_ROOT_MODAL, CLOSE_ROOT_MODAL} from './action_types';
 
 export const openRootModal = () => (dispatch) => {
+    dispatch({
```
marianunez: Nit: could we have this call `openRootModalWithId('')` to avoid code duplication?

scottleedavis:
Note all:  there will be some refactoring on the registry and the programmatic construction of the submenu as requested https://github.com/mattermost/mattermost-webapp/pull/3357#discussion_r319531033

aaronrothschild:
@scottleedavis I'm not on the Repo so I can't remove the "PM Review" Label, sorry.

webapp/src/actions.js
scottleedavis: I agree, though I haven't found a clean way to do it such that dispatch is passed along.  I may be missing something.
```javascript
dispatch(openRootModalWithSubmenuName(''));   //doesn't work either
```

Thoughts on the cleanest way to accomplish this?

scottleedavis:
no worries @aaronrothschild !   thank you.  👍

scottleedavis:
Registry api is updated as part of the MM-16946 code review.  See plugin.jsx.

scottleedavis:
@hanzei the build is failing with the known `unable to access 'https://git.apache.org/thrift.git/`.  Should I include the fix in this PR?

webapp/src/actions.js
scottleedavis: Opposed to leaving this in @marianunez ?  I am open to suggestions on changes, though don't have a better solution currently.

hanzei:
That would be great @scottleedavis. You can copy the solution from https://github.com/mattermost/mattermost-server/pull/12146.

webapp/src/actions.js
marianunez: @scottleedavis I suggest keeping the original `openRootModal` and having the optional text parameter like:

```suggestion
    export const openRootModal = (subMenuText = '') => (dispatch) => {
    dispatch({
        type: SUBMENU,
        subMenu: subMenuText,
    });
    dispatch({
        type: OPEN_ROOT_MODAL,
    });
};
```

I tried it locally and seems to work fine for me. You can ping me in the community server if you want to talk it over.

scottleedavis: This worked great, thank you @marianunez

webapp/src/actions.js
```diff
@@ -1,9 +1,23 @@
...
+        type: SUBMENU,
+        subMenu: subMenuText,
+    });
+    dispatch({
+        type: OPEN_ROOT_MODAL,
+    });
+};
+
+export const openRootModalWithSubmenuName = (text) => (dispatch) => {
```
marianunez: Now that we have `export const openRootModal = (subMenuText = '') => (dispatch)`,  `openRootModalWithSubmenuName` is no longer needed ?

scottleedavis: Apologies for not having removed that.  Done

marianunez:
I added the `Do Not Merge/Awaiting PR` label until we wrap up the related webapp changes and finalize when it will merge. We may need to update the required server version as well.

hanzei:
Just to double this: This will require 5.18, correct?

scottleedavis:
I believe so @hanzei .  
Updated the min_server_version in plugin.json

hanzei:
This can be merged. Sorry for the delay, Scott.

PR #64, opened 2019-09-02
MM-15219 Extend /dialog command to demonstrate introduction text in interactive dialogs

tgkouras:
### Summary
Extended `/dialog` command to demonstrate optional introduction text in the interactive dialogs.

_NOTE_: Mattermost server minimum version will need to be updated appropriately.


server/command_hooks.go
```diff
@@ -355,6 +363,12 @@ func getDialogWithoutElements(state string) model.Dialog {
 	}
 }
 
+func getDialogWithIntroductionText(introductionText string) model.Dialog {
+	dialog := getDialogWithSampleElements()
+	dialog.IntroductionText = introductionText
+	return dialog
```
lieut-data: indentation here seems off?

hanzei: No, it's correct. That just a bug with GitHub.

hanzei:
@tgkouras Please note that you need to update `go.mod` after https://github.com/mattermost/mattermost-server/pull/12029 is merged

hanzei:
@tgkouras After looking at this again, maybe it's a bit to much to add an additional command for this. What do you think about just adding an introduction text to the existing dialogs?

hanzei:
Your're right it's optional and should be threaded as this. Let's leave it as it is.

PR #67, opened 2019-09-13
MM-14724 add an example usage of radio options

patrickkang:
#### Summary
- Add an example usage of radio options
- https://github.com/mattermost/mattermost-server/issues/10465#issuecomment-530852601

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/10465

#### Related Pull Requests
- https://github.com/mattermost/mattermost-webapp/pull/3587


server/command_hooks.go
```diff
@@ -351,6 +351,21 @@ func getDialogWithSampleElements() model.Dialog {
 			Type:        "bool",
 			Placeholder: "Choice:",
 			HelpText:    "Pick a choice",
+		}, {
+			DisplayName: "Radio Option Selector",
+			Name:        "someradiooptionselector",
+			Type:        "radio",
```
hanzei: Would you please set a default element so we can use this for testing of this functionality works?
 ```suggestion
			Type:        "radio",
			Default:     "opt2",
```

prapti:
@patrickkang this mostly works, however Option 1 is selected by default when the interactive dialog opens. If the user wants to keep the selection at Option 1, it doesn't recognize that and still shows the "This field is required" error (please see the screenshot). This needs to either change to not have any default radio option selection like some of the required fields or handled differently.

Screen Shot 2019-09-17 at 8 31 37 AM

PR #68, opened 2019-09-13
[MM-17494] Change the order of translation registration

iomodo:
This PR will change the order of translation registration. 

Fixes https://mattermost.atlassian.net/browse/MM-17494

### DIFF
--- a/webapp/src/plugin.jsx
+++ b/webapp/src/plugin.jsx
@@ -44,6 +44,7 @@ function getTranslations(locale) {
 
 export default class DemoPlugin {
     initialize(registry, store) {
+        registry.registerTranslations(getTranslations);
         registry.registerRootComponent(Root);
         registry.registerPopoverUserAttributesComponent(UserAttributes);
         registry.registerPopoverUserActionsComponent(UserActions);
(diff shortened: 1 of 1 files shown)


webapp/src/plugin.jsx
```diff
@@ -44,6 +44,7 @@ function getTranslations(locale) {
 
 export default class DemoPlugin {
     initialize(registry, store) {
+        registry.registerTranslations(getTranslations);
```
ali-farooq0: shouldn't `registerTranslation` reload components if they're already loaded? if this isn't happening then maybe we should actually be doing that instead rather than relying on the order? unless i'm missing something 🤔

hanzei:
@iomodo Could you please check Ali's feedback?

webapp/src/plugin.jsx
iomodo: Yes it should, it is not a fix to the reload problem. We should move registration forward anyway, why reload translations at all in this case?

ali-farooq0:
@iomodo Can we close this PR if it's not actively being worked? We can turn this into a HW ticket, as I think calling the `registerTranslations` should not depend on when it is called for it to take effect.

iomodo:
Fair enough. Closing the PR

PR #72, opened 2019-10-11
MM-18089 - Custom Admin Settings

marianunez:
#### Summary

Implemented two custom admin settings component that showcases the ability to fully customize the setting area and also provide a setting title that will display the custom setting on the right column of the System Console settings page.

README:

Screen Shot 2019-10-11 at 2 51 57 PM

#### Related PR

[WebApp PR](https://github.com/mattermost/mattermost-webapp/pull/3941)


marianunez:
> Changes look good. My only piece of feedback is wondering if we should change the API to make it explicit which kind of component is being added, and unconditionally render the title in both cases -- on the left side like other components, or as a full width, left-aligned header if not.

After discussing this a bit more online, I've included a `showTitle` flag to indicate that it will show the title and if not present, it will provide the full space to the setting.

webapp/README.md
```diff
@@ -135,7 +135,7 @@ This plugin registers two different admin console settings that are available on
 
 ![custom-setting](docs/admin_setting_full_screen.png)
 
-- `SecretMessage`: Custom setting component that is registerd with a title and is displayed on the right column of the System Console setting page.
+- `SecretMessage`: Custom setting component that is registerd with the option to show title enabled and is displayed on the right column of the System Console setting page.
```
lieut-data:
```suggestion
- `SecretMessage`: Inline, custom setting component shown on the right column of the System Console setting page.
```

webapp/src/plugin.jsx
```diff
@@ -127,7 +127,7 @@ export default class DemoPlugin {
             },
         );
 
-        registry.registerAdminConsoleCustomSetting('SecretMessage', SecretMessageSetting, 'Secret Message');
+        registry.registerAdminConsoleCustomSetting('SecretMessage', SecretMessageSetting, true);
```
lieut-data: Should we accept an options `param` instead?

marianunez: What options did you have in mind?

marianunez: I've wrapped the showTitle in an options object and is passed as an anonymous  object where it clearly indicates the `showTitle` is true.

prapti:
@marianunez Are these the two related Jira tickets for this PR as well (pulled these from the [related webapp](https://github.com/mattermost/mattermost-webapp/pull/3941)) - https://mattermost.atlassian.net/browse/MM-18089, https://mattermost.atlassian.net/browse/MM-19112?

marianunez:
Yes @prapti, those are the webapp implementation to support the custom setting which this PR uses for the demo plugin specifically.

prapti:
@marianunez Does the design aspect fall under the scope of this PR or can that be tested with the web changes separately? When I click on the "Show Secret Message" button, the secret message field is no longer aligned with the right column as the other settings (screenshot no. 1 below).

The other thing is - the custom slider setting is still active when the demo plugin is disabled. Behavior consistency wise, will it make more sense to disable this setting when the plugin is disabled just like the other settings? The other argument to leave it as is, I could assume, is that this setting being custom, demonstrates that it can be left active even when the plugin is disabled, but is it really necessary?
And on a related note, can the slider value be made visible by default (screenshots 2 and 3)? As soon as I slide it, the UI kind of jumps around a little to make room for the slider value.

1. Screen Shot 2019-10-24 at 11 37 09 AM
-----
2. Screen Shot 2019-10-24 at 11 35 54 AM

3. Screen Shot 2019-10-24 at 11 36 05 AM

marianunez:
> Does the design aspect fall under the scope of this PR or can that be tested with the web changes separately? When I click on the "Show Secret Message" button, the secret message field is no longer aligned with the right column as the other settings (screenshot no. 1 below).

@prapti I had missed to push a change 🤦‍♀️that I had locally that should take care of this.  

> The other thing is - the custom slider setting is still active when the demo plugin is disabled.

Correct. It should not be active when the plugin is disabled. Can you double check that you had save the setting after it has set to disabled? It can be confusing that the setting on the top is change and then the Save button must be clicked to take effect. 

> And on a related note, can the slider value be made visible by default (screenshots 2 and 3)? As soon as I slide it, the UI kind of jumps around a little to make room for the slider value.

Yes. Will make this change to fix that.

prapti:
> Correct. It should not be active when the plugin is disabled. Can you double check that you had save the setting after it has set to disabled? It can be confusing that the setting on the top is change and then the Save button must be clicked to take effect.

Thank you for reminding me to Save 😄 
They are disabled once saved. But that is still a little confusing because the pre-existing non-custom settings get enabled/disabled immediately with enabling/disabling the plugin before saving. Should we account for that as well?

marianunez:
> They are disabled once saved. But that is still a little confusing because the pre-existing non-custom settings get enabled/disabled immediately with enabling/disabling the plugin before saving. Should we account for that as well?

Yes, I've updated it with my latest changes to disable the same as the others settings as well. @prapti It should be good now when you retest.

PR #73, opened 2019-11-04
Added /interactive command to demo interactive button functionality

rfoyard:
Related to mattermost/mattermost-server#12697

This is an example to demo the usage of the interactive buttons.  
- Adds an `/interactive` command that display an interactive button
- When the button is clicked, a new post is created containing the request's json.


server/README.md
```diff
@@ -69,15 +68,17 @@ The `/crash` command demonstrates crashing the plugin (and the server recovering
 
 The `/dialog` command demonstrates [Interactive Dialogs](https://docs.mattermost.com/developer/interactive-dialogs.html). Use `/dialog help` for its usage in this demo plugin.
 
-## [http\_hooks.go](http_hooks.go)
+The `/interactive` command demonstrates the usage of interactive buttons.
```
hanzei:
```suggestion
The `/interactive` command demonstrates the usage of interactive message buttons.
```

server/command_hooks.go
```diff
@@ -82,6 +83,15 @@ func (p *Plugin) registerCommands() error {
 		return errors.Wrapf(err, "failed to register %s command", commandTriggerDialog)
 	}
 
+	if err := p.API.RegisterCommand(&model.Command{
+		Trigger:          commandTriggerInteractive,
+		AutoComplete:     true,
+		AutoCompleteHint: "",
+		AutoCompleteDesc: "Interactive button demonstration",
```
hanzei:
```suggestion
		AutoCompleteDesc: "Demonstrates  interactive message buttons.",
```

server/command_hooks.go
```diff
@@ -405,6 +417,39 @@ func getDialogWithIntroductionText(introductionText string) model.Dialog {
...
+
+	post := &model.Post{
+		ChannelId: args.ChannelId,
+		UserId:    p.botId,
+		Message:   "Test interactive button",
+		Props: model.StringInterface{
+			"attachments": []*model.SlackAttachment{{
+				Actions: []*model.PostAction{{
+					Integration: &model.PostActionIntegration{
+						URL: fmt.Sprintf("%s/plugins/%s/interaction/action", siteURL, manifest.Id),
```
hanzei: The `siteURL` is actually not longer needed.

```suggestion
						URL: fmt.Sprintf("/plugins/%s/interaction/action", manifest.Id),
```

hanzei: Maybe went should use a more specific URL here. Something like e.g. `interactive/button/1`.

server/http_hooks.go
```diff
@@ -229,6 +231,37 @@ func (p *Plugin) handleEphemeralDelete(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	user, appErr := p.API.GetUser(request.UserId)
+	if appErr != nil {
+		p.API.LogError("failed to get user for interaction action", "err", appErr.Error())
+		w.WriteHeader(http.StatusOK)
+		return
+	}
+
+	requestJSON, _ := json.MarshalIndent(request, "", "  ")
```
hanzei: Please check the error.

rfoyard: Oops, copied this from line 54 and did not modify thereafter, will do.

rfoyard:
Thank you @hanzei

hanzei:
Thanks _you_ for the PR!

prapti:
@rfoyard Could you please add some test steps in the description?
@hanzei There is no Jira ticket for this particular PR but would https://mattermost.atlassian.net/browse/MM-10164 serve as the ticket for this PR as well?

hanzei:
@prapti Yes, we can use this JIRA ticket.

prapti:
@rfoyard Thank you!
A couple questions:
1. Clicking on the button generates a response that is grouped together in the reply thread (this is great!) but the reply count on the reply button does not get updated.
2. Issuing a `/interactive` command on the RHS reply thread generates the button in the main center channel. Is this by design? Would it be possible to generate the button on the RHS?
Since the interactive button itself does what is expected, we could merge this one and open improvement tickets if the issues are out of scope for this PR.
@hanzei @marianunez are there any thoughts/comments on these questions?

rfoyard:
> 1. Clicking on the button generates a response that is grouped together in the reply thread (this is great!) but the reply count on the reply button does not get updated.

This is the same behaviour as the `/dialog` command.  
I am not sure how the ReplyCount is updated nor if it's even possible for bots' posts.

> 2. Issuing a `/interactive` command on the RHS reply thread generates the button in the main center channel. Is this by design? Would it be possible to generate the button on the RHS?

For this one, it's the same behaviour as the `/ephemeral` command.  
I just tried to create the button in the RHS, but I could not get the response to be grouped together (neither in the center channel nor in the RHS)

server/command_hooks.go
```diff
@@ -405,6 +417,38 @@ func getDialogWithIntroductionText(introductionText string) model.Dialog {
 	return dialog
 }
 
+func (p *Plugin) executeCommandInteractive(args *model.CommandArgs) *model.CommandResponse {
+
+	post := &model.Post{
+		ChannelId: args.ChannelId,
+		UserId:    p.botId,
```
hanzei: Please add to fix the second issue
```suggestion
		RootId:    args.RootId,
		UserId:    p.botId,
```

server/command_hooks.go
```diff
@@ -405,6 +417,38 @@ func getDialogWithIntroductionText(introductionText string) model.Dialog {
 	return dialog
 }
 
+func (p *Plugin) executeCommandInteractive(args *model.CommandArgs) *model.CommandResponse {
+
+	post := &model.Post{
+		ChannelId: args.ChannelId,
```
hanzei: Please add this to fix the second issue
```suggestion
		ChannelId: args.ChannelId,
		RootId:    args.RootId,
```

rfoyard: I already tried this, but then, clicking the button will trigger an error in `http_hooks.go`: `createPost: Invalid RootId parameter`. And if I remove the `RootId`, the grouping is not respected anymore.

hanzei: Could you please double check? My code change is about the post created via the slash command, not when the button is pressed.

rfoyard: I know, but in `http_hooks.go` I use the `request.PostId` as the new Post's RootId and when I change your line, it fails...

hanzei: > I know, but in `http_hooks.go` I use the `request.PostId` as the new Post's RootId and when I change your line, it fails...

I see. But this is actually not correct. If the post with the buttons is not a root post, we should not use it's id as RootId. Instead we should get the post with `request.PostId` and use **it's** rootId.

rfoyard: Thanks, it fixed the issue !

server/http_hooks.go
```diff
@@ -229,6 +231,51 @@ func (p *Plugin) handleEphemeralDelete(w http.ResponseWriter, r *http.Request) {
...
+		p.API.LogError("failed to marshal json for interactive action", "err", jsonErr.Error())
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	msg := "@%v clicked an interactive button.\n```json\n%v\n```"
+	if _, appErr := p.API.CreatePost(&model.Post{
+		UserId:    p.botId,
+		ChannelId: request.ChannelId,
+		RootId:    post.RootId,
```
hanzei: If `post.RootId` is `""`, we need to use `post.Id` here.

hanzei:
@prapti Please take another look

hanzei:
@rfoyard Can you please merge `master` into your branch? The changes to `go.*` are not longer needed.

hanzei:
Thanks for this great addition to the demo plugin :+1:

rfoyard:
> Thanks for this great addition to the demo plugin 👍
You're very welcome !

PR #83, opened 2020-01-13
MM-21683: expand boolean defaults for testing

lieut-data:
#### Summary
Add new boolean interactive dialog elements to test the various defaults.

#### Ticket Link
Relates-to: MM-21683


lieut-data:
Good call, @prapti! What do you think of this instead?  
image

PR #84, opened 2020-01-26
Add more manifest metadata

hanzei:
#### Summary
Add `homepage_url` and `support_url` to the manifest. `support_url` is not jet used in the webapp, but there is no downside of defining it already.

I also added an icon for the marketplace. 
Screenshot from 2020-01-27 20-24-54

I've tested the changes locally and can confirm that the link to the `homepage_url` is correctly shown and the icon is shown.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-21918


hanzei:
@asaadmahmood What icon should be use for the demo plugin? Maybe just the Mattermost icon?

PR #85, opened 2020-02-03
MM-21984: introduce background job

lieut-data:
#### Summary
Depend on the experimental [github.com/lieut-data/mattermost-plugin-api/cluster](https://pkg.go.dev/github.com/lieut-data/mattermost-plugin-api/cluster?tab=doc) package to spawn a background job that coordinates among multiple plugin instances automatically.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-21984


go.sum
```diff
@@ -197,6 +202,9 @@ github.com/mattermost/gorp v2.0.1-0.20190301154413-3b31e9a39d05+incompatible h1:
 github.com/mattermost/gorp v2.0.1-0.20190301154413-3b31e9a39d05+incompatible/go.mod h1:0kX1qa3DOpaPJyOdMLeo7TcBN0QmUszj9a/VygOhDe0=
 github.com/mattermost/ldap v3.0.4+incompatible h1:SOeNnz+JNR+foQ3yHkYqijb9MLPhXN2BZP/PdX23VDU=
 github.com/mattermost/ldap v3.0.4+incompatible/go.mod h1:b4reDCcGpBxJ4WX0f224KFY+OR0npin7or7EFpeIko4=
+github.com/mattermost/mattermost-server v5.11.1+incompatible h1:LPzKY0+2Tic/ik67qIg6VrydRCgxNXZQXOeaiJ2rMBY=
+github.com/mattermost/mattermost-server v5.11.1+incompatible/go.mod h1:5L6MjAec+XXQwMIt791Ganu45GKsSiM+I0tLR9wUj8Y=
```
hanzei: I'm surprised to see this here. See my comment at https://github.com/lieut-data/mattermost-plugin-api/commit/3ffb8392917af674c7fe0b47477259433b9211f6#r37086166

server/activate_hooks.go
```diff
@@ -56,6 +59,17 @@ func (p *Plugin) OnActivate() error {
 		}
 	}
 
+	job, cronErr := cluster.Schedule(
```
hanzei: I quite like the API. :+1: 

I'm wondering if the makes sense to have the `API` as the first parameter, similar how `ctx` should always be the first argument. It makes sense to me to put the parameters that are most of the time the same as the first ones.

server/activate_hooks.go
```diff
@@ -56,6 +59,17 @@ func (p *Plugin) OnActivate() error {
...
+		"BackgroundJob",
+		cluster.JobConfig{Interval: 1 * time.Minute},
+		p.BackgroundJob,
+		p.API,
+	)
+	if cronErr != nil {
+		p.API.LogError("Failed to schedule background job", "err", cronErr)
```
hanzei: Should the code return here?

server/activate_hooks_test.go
```diff
@@ -77,8 +77,11 @@ func TestOnActivate(t *testing.T) {
 			},
 			ShouldError: true,
 		},
-		"minimum supported version fullfiled": {
+		"minimum supported version fulfilled": {
 			SetupAPI: func(api *plugintest.API) *plugintest.API {
+				api.On("KVSetWithOptions", mock.Anything, mock.Anything, mock.AnythingOfType("model.PluginKVSetOptions")).Return(true, nil).Maybe()
+				api.On("KVGet", mock.Anything).Return(nil, nil).Maybe()
```
hanzei: That is a to me the most biggest downside of the design: When using mocking for tests, the plugins has to know the implementation of the helper.

@levb I've seen some effort to split plugins into multiple packages. Do you think that approach can help getting rid of mocking?

lieut-data: Yeah, I intentionally did the minimal amount of work necessary to get the tests to pass here. I think the right place to "fix" this in the plugin API rework, though I admit exactly how is still unknown. Just anything other than this ;P

server/activate_hooks.go
lieut-data: Ah, yeah, for the demo plugin, that makes sense. Good call!

go.sum
lieut-data: Good catch, definitely unintentional on my part! I've pushed a `go mod tidy` to the various repositories here.

hanzei:
@lieut-data Heads up that go reports a data race when running the tests for the `cluster` package: https://github.com/lieut-data/mattermost-plugin-api/runs/423102088

server/activate_hooks_test.go
levb: @hanzei We are moving towards wrapping all p.API dependencies in a consistent (for our purposes), plugin-specific interfaces. In the case of MS Calendar (WIP) you can see the implementation in https://github.com/levb/mattermost-plugin-mscalendar/blob/c8d6579d3324c5802c4f83c6533dd1c56c5ee614/server/mscalendar/mscalendar.go#L32-L38 and https://github.com/levb/mattermost-plugin-mscalendar/blob/lev-package-changes/server/utils/pluginapi/api.go.

The former defines the interface needed by MSCalendar; the latter provides a shareable "base" implementation, a go Type with a lot of methods. Then https://github.com/levb/mattermost-plugin-mscalendar/blob/c8d6579d3324c5802c4f83c6533dd1c56c5ee614/server/plugin/plugin.go#L232 ties them altogether.

Also, https://github.com/levb/mattermost-plugin-mscalendar/tree/lev-package-changes/server/utils/bot for all logs, messages, eventually webhooks etc. https://github.com/levb/mattermost-plugin-mscalendar/tree/lev-package-changes/server/utils is intended to be shareable, does not have any mscalendar dependencies.

This has been a result of slow evolution of our work, we are starting to converge across `workflow`, `mscalendar`, `autolink`, and `solar-lottery`. This layer may more useful for the more complex integration, and not intended as a replacement for the underlying "base" plugin API.

These specific interfaces are then easy to mock, as needed, and we use `gomock` for them.

cc @crspeller

server/activate_hooks.go
lieut-data: Agreed!

lieut-data:
Thanks, @hanzei! Looks like my test code was missing a synchronization point. I've updated to v0.0.6 with those changes as well as the tweak to the parameter order as suggested.

lieut-data:
@crspeller, I'd love to use `error` in those interfaces, but I don't think I can without some kind of translation layer to wrap `p.API` accordingly. In fact, that's precisely what the parent repository (https://pkg.go.dev/github.com/lieut-data/mattermost-plugin-api@v0.0.6?tab=doc) purports to do, and what I imagine the `cluster` package will eventually consume instead of the RPC API directly. 

This first pass is mostly meant to unblock ongoing efforts without having to opt-in to the larger (and incomplete) plugin API rework.

server/plugin.go
```diff
@@ -18,4 +19,8 @@ type Plugin struct {
 
 	// BotId of the created bot account.
 	botId string
+
+	// backgroundJob is a job that executes periodically on only one plugin instance at a time
+	// interval
```
lieut-data:
```suggestion
```

lieut-data:
I've discovered that https://github.com/mattermost/mattermost-server/pull/13757 is required for the lock to behave correctly in cases where the lock isn't explicitly deleted. Without that change, `KVCompareAndSet` doesn't honour expiry at all.

lieut-data:
Good questions, @levb! Lots of opportunities to improve the documentation here as well.

> 1. As I understood it, the job mutex is good for 15 seconds, otherwise needs to be renewed, but the renewal is not currently coded in job.go. Should we be clear that jobs MUST complete in 15 seconds or renew the lock themselves?

The renewal logic is internal to the mutex, and will be automatically refreshed until `Unlock()` is invoked. If the plugin instance dies, taking the refresh job along with it, the corresponding key value will expire, unlocking the mutex automatically. 

The job is free to run as long as necessary, but the strategy to deal with a half-completed job from a dead plugin instance is left to the plugin. Ideally, jobs would be idempotent and the next scheduled instance would seamlessly pick up / restart from scratch.

> 2. Config changes - what exactly is the flow of handling a configuration change if my job is already running. It doesn't appear like killing/re-creating the job is the right thing to do, but maybe as a goroutine, so that the current one finishes? Maybe an example?

At the moment, the only configuration change that's relevant to the /package/ is the scheduled interval. Right now, this is fixed at the time the job is created. It would be necessary to `job.Close()` (which would block until the current job is done) and then reschedule the job to change this interval. Each plugin instance would have to do this for the interval to matter (no need to explicitly coordinate, just that if they aren't in sync, the job will be run as often as the smallest interval configured).

The configuration semantics of the actual work to be done by the job can change however the plugin sees fit. It's probably the case that access to the configuration parameters should be synchronized, i.e. `p.getConfiguration()` call followed by working from that copy of the config.

But perhaps one thing that's missing is a means to interrupt the current job for any reason. It's possible to build this "outside" the package, but it might be convenient for the callback to accept a `context.Context` and to be able to configure or access this from outside the job. I'm inclined to defer this until a concrete use-case arises.

> 3. I did not fully understand the appError vs error response - why again can't the job interface be all defined as error even if it's just an underlying *model.AppError being passed? It's not a blocker IMO but then should we just say no to new AppErrors in all new package interfaces?

The `Schedule` and `NewMutex` APIs are currently designed to consume `p.API` directly, e.g. to receive a method named `KVGet(key string) ([]byte, *model.AppError)`. We could absolutely change the interface to accept a `KVGet(key string) ([]byte, error)` instead, but it would no longer accept `p.API` directly. 

Instead, we would need a complementary struct that wraps `p.API` and implements this interface, e.g.:
```go
type APIWrapper struct {
    pluginAPI plugin.API
}

func (w *APIWrapper) KVGet(key string) ([]byte, error) {
    value, appErr := w.pluginAPI.KVGet(key)
    if appErr != nil {
        return nil, appErr
    }

    // Can't just return value, appErr, since a `nil` appErr does not a `nil` error make.
    return value, nil
}
```

Since the plan remains to fix "all of this" holistically -- as partially implemented in the parent repository -- /and/ we're free to break this interface as we see fit (i.e. follow semver at the package level without any direct coupling to mattermost-server), it seems to make sense to optimize for ease-of-use right now.

hanzei:
> We could absolutely change the interface to accept a `KVGet(key string) ([]byte, error)` instead, but it would no longer accept `p.API` directly.

Maybe I totally missing the point here, but `model.AppError` implements `error`. Shouldn't `a.API` then also fulfill the interface with `KVGet(key string) ([]byte, error)` ?

lieut-data:
@hanzei, I agree that might be convenient, but I don't believe interfaces afford that flexibility. See https://play.golang.org/p/KJM4T0-ezJu, but:
```go
package main

type AppError struct {
}

func (appError *AppError) Error() string {
    return "error"
}

type apiInterface interface {
    KVGet(key string) ([]byte, error)
}

type api struct {
}

func (p *api) KVGet(key string) ([]byte, *AppError) {
    return nil, nil
}


func Schedule(api apiInterface) {
}

func main() {
	api := &api{}
	Schedule(api)
}
```

fails with:
```
./prog.go:27:10: cannot use api (type *api) as type apiInterface in argument to Schedule:
	*api does not implement apiInterface (wrong type for KVGet method)
		have KVGet(string) ([]byte, *AppError)
		want KVGet(string) ([]byte, error)
```

levb:
> Since the plan remains to fix "all of this" holistically.

@lieut-data I'm good on 1 and 2, and I learned to believe in master plans while growing up in the Soviet Union, so 3 is good, too :)

lieut-data:
This is pending review of https://github.com/lieut-data/mattermost-plugin-api/pull/33, after which we can update our dependencies here and merge.

prapti:
@lieut-data I might need your help understanding/testing this PR. Would it be possible to add some initial test steps in Jira?

lieut-data:
Thanks for the reminder, @prapti! Test steps added.

lieut-data:
Heads up that I've changed the default interval to once every 15 minutes to avoid spamming the channel too much while the demo plugin is enabled. Otherwise, the behaviour remains unchanged.

PR #89, opened 2020-02-11
MM-22414: relax required server configuration

lieut-data:
#### Summary
Instead of hard-failing the demo plugin, just log an error when
showcasing the "required configuration" portion of the plugin. Right
now, the error message doesn't specifically call out next steps for
someone using the plugin, making it hard to use as a feature.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-22414


hanzei:
@lieut-data Do we need to cut a new release with these changes?

lieut-data:
@hanzei, possibly, though I imagine no hurry this week with MatterCon. Maybe we can squeeze in the background job PR while we're at it.

PR #90, opened 2020-03-05
MM-21987: Demonstrate resolution of mentions

agarciamontoro:
#### Summary

This PR adds a `/show_mentions` command that showcases the new functionality added in
https://github.com/mattermost/mattermost-server/pull/13762, which implements a feature to resolve mentions in slash commands, linking each mention to team users or public channels to their respective IDs.

The demo plugin answers to the new command with a pair of Markdown tables that contain, respectively, the mentions to team members and mentions to public channels, as shown in the following screenshot:

Screenshot of new command

Opening as Draft PR because https://github.com/mattermost/mattermost-server/pull/13762 is not merged. Once it is, I will update go.mod to point the dependency on mattermost-server to its new `HEAD`.

If you want to test it locally, make sure that you have the MM-21987 branch from https://github.com/mattermost/mattermost-server/pull/13762 in your local direectory and use a replace directive in the `go.mod`, as in
```bash
go mod edit -replace github.com/mattermost/mattermost-server/v5=/path/to/your/local/mattermost-server@MM-21987
```

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-21987


agarciamontoro:
@hanzei @crspeller I requested the review to the first two people that Github suggested, so feel free to unassign yourselves and I'll bug other people :)

agarciamontoro:
Oops, there are some conflicting files. Let me fix those conflicts first.

server/command_hooks.go
```diff
@@ -522,6 +534,50 @@ func (p *Plugin) crash() {
 	_ = 1 / y
 }
 
+func (p *Plugin) executeCommandMentions(args *model.CommandArgs) *model.CommandResponse {
+	message := "The command `" + args.Command + "` contains the following different mentions.\n"
```
hanzei: Does it make sense to make use of `text/template` here and get rid of the string concatenation?

server/command_hooks.go
```diff
@@ -522,6 +534,50 @@ func (p *Plugin) crash() {
...
+	message += "### Mentions to users in the team\n"
+	if args.UserMentions == nil {
+		message += "_There are no mentions to users in the team in your command_.\n"
+	} else {
+		message += "| User name | Unique identifier |\n"
```
hanzei: Nit: `ID` is the commonly used at Mattermost 
```suggestion
		message += "| User name | ID |\n"
```

server/command_hooks.go
```diff
@@ -522,6 +534,50 @@ func (p *Plugin) crash() {
...
+		message += "|-----------|-------------------|\n"
+		for name, id := range args.UserMentions {
+			message += fmt.Sprintf("|@%s|%s|\n", name, id)
+		}
+	}
+
+	message += "\n### Mentions to public channels\n"
+	if args.ChannelMentions == nil {
+		message += "_There are no mentions to public channels in your command_.\n"
```
hanzei: From the field name it isn't clear to me that `args.ChannelMentions` only contains mentions to public channels. Could you please elaborate a bit why only mentions to public channels are resolved and why the name is `ChannelMentions`?

agarciamontoro: The rationale behind resolving only public channels comes from the feature proposal (https://mattermost.atlassian.net/browse/MM-21987). @lieut-data, could you elaborate a bit on why we decided to resolve only public channels? I just assumed it at the moment and did not ask this to you.
In any case, @hanzei is right that it would have been more sensible to call this `PublicChannelMentions`, instead. We should open a ticket to track this, it may be a bit confusing for the end user.

server/command_hooks.go
agarciamontoro: Did not know about that package, I'll make use of it, thanks!

go.sum
```diff
@@ -195,22 +221,38 @@ github.com/mattermost/go-i18n v1.11.0 h1:1hLKqn/ZvhZ80OekjVPGYcCrBfMz+YxNNgqS+be
...
 github.com/mattermost/gorp v2.0.1-0.20190301154413-3b31e9a39d05+incompatible/go.mod h1:0kX1qa3DOpaPJyOdMLeo7TcBN0QmUszj9a/VygOhDe0=
+github.com/mattermost/gosaml2 v0.3.2/go.mod h1:Z429EIOiEi9kbq6yHoApfzlcXpa6dzRDc6pO+Vy2Ksk=
+github.com/mattermost/ldap v0.0.0-20191128190019-9f62ba4b8d4d h1:2DV7VIlEv6J5R5o6tUcb3ZMKJYeeZuWZL7Rv1m23TgQ=
+github.com/mattermost/ldap v0.0.0-20191128190019-9f62ba4b8d4d/go.mod h1:HLbgMEI5K131jpxGazJ97AxfPDt31osq36YS1oxFQPQ=
 github.com/mattermost/ldap v3.0.4+incompatible h1:SOeNnz+JNR+foQ3yHkYqijb9MLPhXN2BZP/PdX23VDU=
 github.com/mattermost/ldap v3.0.4+incompatible/go.mod h1:b4reDCcGpBxJ4WX0f224KFY+OR0npin7or7EFpeIko4=
+github.com/mattermost/mattermost-server/v5 v5.3.2-0.20200311105012-2bec92a4040f h1:K5q4qlurk4CjHP//qbftYSW1yKUjoekOFT++sbab3EQ=
+github.com/mattermost/mattermost-server/v5 v5.3.2-0.20200311105012-2bec92a4040f/go.mod h1:mRbRNCKoD3iN1TYjolNjBO+U8/SA3JLOo3Q4IONfmh0=
 github.com/mattermost/mattermost-server/v5 v5.18.0 h1:In1v/1vrYMEDRHiXe6sA50KJ5WGkE2mLtsIz7YOGsN8=
 github.com/mattermost/mattermost-server/v5 v5.18.0/go.mod h1:10vIYDohcPNDMDk3XlRUjNO9nGk9fnFHEFeSIUYR82k=
```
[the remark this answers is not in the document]
agarciamontoro: Oops, nope, I'll do it, thanks for the tip!

server/command_hooks.go
lieut-data: 3/5, but importing `text/template` just to get rid of string concatenation seems like overkill.

server/command_hooks.go
lieut-data: My thinking here was to match the webapp functionality -- private channels are never autocompleted, and only show as "mentions" if you happen to be in the channel. I made the assumption that we should start with privacy here, and expand later if we decide there's no privacy concerns.

server/command_hooks.go
agarciamontoro: I first thought of that, but nevertheless tried to implement it with a template to see how it looked like. However, I have now another counterargument: the goal of this command is to showcase how to use the new structures in the code itself, and inlining the access to the mentions structures inside the template, makes it harder to read.
So I would also leave it as it is right now. Further thoughts, @hanzei?

server/command_hooks.go
agarciamontoro: Thank you, makes total sense. In any case, I think we can make that change to the name of the field before it's released.

server/command_hooks.go
agarciamontoro: This is powerful (thanks for the tip!) but hides all the important details in this specific case:
```go
	const message = `
The command` + "`{{.Command}}`" + `contains the following different mentions.

### Mentions to users in the team
{{if not .UserMentions}}
_There are no mentions to users in the team in your command_.
{{else}}
| User name | ID |
|-----------|----|
{{- range $name, $id := .UserMentions}}
|@{{$name}}|{{$id}}|
{{end}}
{{end}}

### Mentions to public channels
{{if not .ChannelMentions}}
_There are no mentions to public channels in your command_.
{{else}}
| Channel name | ID |
|--------------|----|
{{range $name, $id := .ChannelMentions -}}
|~{{$name}}|{{$id}}|
{{end}}
{{end}}
`
```

go.sum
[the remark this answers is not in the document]
agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-demo/pull/90/commits/ade73fb020d10b8a34a2cd2435df811b7c2b0e9e

server/command_hooks.go
agarciamontoro: Ticket opened to document the whole structure: https://mattermost.atlassian.net/browse/MM-23268. I'd mark this comment as resolved, as long as @hanzei agrees.

server/command_hooks.go
hanzei: Makes sense :+1: I'm fine leaving it as it is

server/command_hooks.go
hanzei: The naming makes sense to me :+1: 

FYI: https://docs.mattermost.com/developer/slash-commands.html#custom-slash-command will also need an update.

PR #91, opened 2020-03-09
Sync with plugin-starter-template

hanzei:
#### Summary
Downstream changes from https://github.com/mattermost/mattermost-plugin-starter-template


plugin.json
```diff
@@ -78,7 +78,7 @@
             "display_name": "Secret Message",
             "type": "custom",
             "help_text": "The message posted by the demo plugin when the secret phrase is detected.",
-            "default": "Yay! The random secret string was posted!\n\nGo to the settings page for this plugin in the system console to generate a new random secret."
+            "default": "Yay! The random secret string was posted! Go to the settings page for this plugin in the system console to generate a new random secret."
```
hanzei: I had to change this because `\n` did play nicely with the js world.

server/activate_hooks.go
```diff
@@ -96,15 +94,5 @@ func (p *Plugin) OnDeactivate() error {
...
-	path, err := p.API.GetBundlePath()
-	if err != nil {
-		return false, errors.Wrap(err, "failed to get bundle path")
-	}
-
-	manifest, _, err := model.FindManifest(path)
-	if err != nil {
-		return false, errors.Wrap(err, "failed to find manifest file")
-	}
-
```
marianunez: Is this intentionally deleted?

hanzei: Yes. The method can reuse https://github.com/mattermost/mattermost-plugin-demo/pull/91/files#diff-3407a86002fad9e52eb18c8919aa27b5R11 and doesn't have to fetch the manifest itself

build/manifest/main.go
```diff
@@ -10,19 +10,36 @@ import (
...
-
-var manifest = struct {
-	Id      string
-	Version string
-}{
-	Id:      "%s",
-	Version: "%s",
+const pluginIDGoFileTemplate = `// This file is automatically generated. Do not modify it manually.
+
+package main
```
jfrerich: @hanzei is was there some funky merging here?  package main defined more than once..

hanzei:
Let's wait for https://github.com/mattermost/mattermost-plugin-starter-template/pull/89

build/manifest/main.go
hanzei: I'm not sure I'm getting was the issue is. The generated file gets placed into `server`, which contains all needed `.go` files for the plugin to build. They are all in `package main`.

jfrerich: I was thrown off by the additional `package main` line while skimming through the files and didn't see that is included as a string.

hanzei:
@DHaussermann Would you please give this PR a smoke test?

hanzei:
Let's wait and cut `v0.6.0` first.

PR #93, opened 2020-03-12
MM-23227: preserve disable across config changes

lieut-data:
#### Summary
Fix an issue with how the configuration structure was always being reinitialized whenever a system configuration change occurred. Also avoid mutating the configuration in place.

Note that the `go mod tidy` commit was run on vanilla master, not specifically my changes. Not sure how that slipped in.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23227


lieut-data:
Oops, yeah -- looks like the tests were actually failing. Fixed!

PR #102, opened 2020-05-04
MM-24503 Adding custom route example.

crspeller:
#### Summary
Adding example of route registration added in: https://github.com/mattermost/mattermost-webapp/pull/5367

Routes can be found at:
- `/plug/com.mattermost.demo-plugin/roottest`
- `/<teamname>/com.mattermost.demo-plugin/teamtest`

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24503

### DIFF
--- a/webapp/README.md
+++ b/webapp/README.md
@@ -145,3 +145,9 @@ This plugin registers two different admin console settings that are available on
 
 ![secret-message-setting](docs/admin_setting_with_title.png)
 
+## Custom routes
+
+This plugin registers two custom routes. 
+
+- `/plug/com.mattermost.demo-plugin/roottest` which brings you to a page with the text `Demo plugin route.`
+- `/<teamname>/com.mattermost.demo-plugin/teamtest` which brings you to a page with the text `Demo plugin team route.`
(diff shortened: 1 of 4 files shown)


crspeller:
@hanzei Thanks! I had no idea that file existed.

marianunez:
Could we add a link to the new routes? In the root modal or the RHS possibly so that they are easily accesible within the plugin UI?

lieut-data:
Agree with @marianunez -- I think it's important that the links are in the main webapp, so we can also verify that navigation occurs without refreshing the page.

crspeller:
@marianunez @lieut-data Ya that makes a lot of sense. An example of how to make it not refresh is a good idea. 

Added some links to the RHS.

prapti:
@crspeller While I'm going to review this PR (and the associated web PR) after @cpoile's review is completed, I was wondering that to test that there was no page refresh, verifying that no HTTP requests was sent when clicking that link on the demo-RHS or following that link from the address bar would be a way to test it.

crspeller:
@prapti Yes that sounds reasonable.

PR #123, opened 2020-10-11
Adds demonstration of reaction hooks

shieldsjared:
#### Summary
Demonstrating the use of the proposed/pending `ReactionHasBeenAdded` and `ReactionHasBeenRemoved` plugin hooks. Temporarily depends on [forked repository](https://github.com/shieldsjared/mattermost-server/tree/GH-15624) of `mattermost-server` until the [PR for these hooks](https://github.com/mattermost/mattermost-server/pull/15765) is approved/merged.

The demonstration follows the pattern of the other hooks.  Let me know if you'd like something a little more creative or if this sufficient for the demo plugin :) 

Also, the code I added for each of the hooks is highly duplicative; however, I noticed some of the other demonstration hooks are fairly self-sufficient as well, so I followed suit (perhaps to keep demo plugin easy to read/follow?).  If you'd prefer the duplicate code to be extracted, I'm happy to adjust it!

Lastly, I snuck in one variable name change in `postPluginMessage()` for clarity.

#### Ticket Link
GH Issue: https://github.com/mattermost/mattermost-server/issues/15624
Pull Request: https://github.com/mattermost/mattermost-server/pull/15765


server/plugin_message.go
```diff
@@ -6,7 +6,7 @@ import (
 	"github.com/mattermost/mattermost-server/v5/model"
 )
 
-func (p *Plugin) postPluginMessage(id, msg string) *model.AppError {
+func (p *Plugin) postPluginMessage(teamID, msg string) *model.AppError {
```
levb, line 9: 👍

levb (approved):

go.mod
```diff
@@ -8,5 +8,7 @@ require (
 	github.com/mattermost/mattermost-server/v5 v5.25.1
 	github.com/mholt/archiver/v3 v3.3.0
 	github.com/pkg/errors v0.9.1
-	github.com/stretchr/testify v1.5.1
+	github.com/stretchr/testify v1.6.1
 )
+
+replace github.com/mattermost/mattermost-server/v5 v5.25.1 => github.com/shieldsjared/mattermost-server/v5 v5.0.0-20201008141002-56ee0d27d54b
```
hanzei, line 14: Good use of `replace` :+1:

server/reaction_hooks.go
```diff
@@ -0,0 +1,108 @@
...
+func (p *Plugin) ReactionHasBeenAdded(c *plugin.Context, reaction *model.Reaction) {
+	configuration := p.getConfiguration()
+
+	if configuration.disabled {
+		return
+	}
+
+	user, err := p.API.GetUser(reaction.UserId)
+	if err != nil {
+		p.API.LogError("failed to query user", "user_id", reaction.UserId)
```
hanzei, line 25: Please also log `err` to allow easier debugging

server/reaction_hooks.go
```diff
@@ -0,0 +1,108 @@
...
+		return
+	}
+
+	post, err := p.API.GetPost(reaction.PostId)
+	if err != nil {
+		p.API.LogError("failed to query post", "post_id", reaction.PostId)
```
hanzei, line 31: Nit pick: Usually log messages are capitalized 
```suggestion
		p.API.LogError("Failed to query post", "post_id", reaction.PostId)
```

server/reaction_hooks.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		p.API.LogError("failed to query channel", "channel_id", post.ChannelId)
+		return
+	}
+
+	team, err := p.API.GetTeam(channel.TeamId)
+	if err != nil {
+		p.API.LogError("failed to query team", "team_id", channel.TeamId)
+		return
+	}
```
hanzei, line 45: If the emojis is removing from a post in a DM, this call will fail. You can use construct the link using `_redirect/pl/$POST_ID` similar to https://github.com/matterpoll/matterpoll/blob/45f095875a98fb1d4f3f166851c86f41b987493e/server/plugin/api.go#L665

hanzei (changes requested):
Nice work, the demonstration is creative enough :wink:

server/reaction_hooks.go
shieldsjared: @hanzei - its lowercase throughout this plugin repo. I can go ahead and make it capitalized in the other files too if you're OK w/ that being bundled w/ this PR (so it all remains consistent).

server/reaction_hooks.go
shieldsjared: @hanzei - Not currently included in `channel_hooks.go`, `message_hooks.go`, `team_hooks.go`. Want me to update those in this PR as well to be consistent? Some folks prefer _pure_ PR's, other's are OK w/ grouping minor things like that. Wasn't sure on your preference.

server/reaction_hooks.go
shieldsjared: @hanzei - Ah, great insight, thank you! Didn't realize that as another way to construct the link.

One thought though... I just tested a direct message, and the code actually fails before ever getting to this point.  Specifically the call before it to `GetChannel(post.ChannelId)`  fails for a direct message, so the demo plugin never gets to log the event.  I checked and it looks like this is also happening in the current setup for `MessageHasBeenPosted` (e.g. if you post the plugin secret in a direct message, it will fail and not log to the demo channel)... Potentially other places too?

As written, these are two sets of functionality (in the context of this plugin) that fail during direct messages.  Do we want to figure out a way to adapt them to work for DM's (if so, I can change the URL above)... Or do we want to leave them as-is, knowing they will fail for DM's?

server/reaction_hooks.go
hanzei: Yes, logging the error should become good practise. I don't mind of you change the other files in this PR too, but that is not an requirement.

server/reaction_hooks.go
hanzei: Good point. I didn't know that. Consistent is key here. If you want, you can change all instances, if not it can stay as it is.

server/reaction_hooks.go
hanzei: Are you sure? For me it fails with `failed to query team` while the channel gets fetched correctly. This is also the behaviour that I was expecting.

server/reaction_hooks.go
shieldsjared: Sounds good! Went ahead and updated anywhere I else I noticed it was missing

server/reaction_hooks.go
shieldsjared: Went ahead and changed all `LogError` and `LogWarn` instances that I found, so everything's consistent.

server/reaction_hooks.go
shieldsjared: Whoops, you're correct, I got ahead of myself and tested after I changed the code.  However, the overall hook still fails to post the message to the demo channel, since there's no team to correlate it to.

I went ahead and updated the URL to the pattern you mentioned.

We're still left with `postPluginMessage()` failing when called for a post that was from a DM (so you'll never see the logged message for a DM).  Do you want to accept that as-designed (it currently impacts the message hooks too)... or do you want to consider options for how to keep the functionality when executed from a DM?  For example, when the correct demo channel cannot be found, we could always use the bot to direct message the log message back to the initiator...

server/reaction_hooks.go
```diff
@@ -0,0 +1,118 @@
...
+			"Failed to query channel",
+			"channel_id", post.ChannelId,
+			"error", err.Error(),
+		)
+		return
+	}
+
+	postURL := fmt.Sprintf("%s/_redirect/pl/%s", *p.API.GetConfig().ServiceSettings.SiteURL, reaction.PostId)
+	msg := fmt.Sprintf("ReactionHasBeenRemoved: @%s, :%s:, [<jump to convo>](%s)", user.Username, reaction.EmojiName, postURL)
+	if err := p.postPluginMessage(channel.TeamId, msg); err != nil {
```
hanzei, line 110: There is still an issue with DM's. The plugin has a list of `demoChannelIds` which represent the demo channels per team. If `channel.TeamId` is empty i.e. `channel` is a DM, https://github.com/mattermost/mattermost-plugin-demo/blob/5f04050e3c54f44606fa9f8e47dd304774c17e76/server/plugin_message.go#L23 will be empty and posting the message will fail. I think it's fine to post the message into every team for DMs.

hanzei (changes requested):
Thanks for the cleanup :+1: 

There is still one issue with DMs.

server/reaction_hooks.go
hanzei: @shieldsjared do you have any questions about the feedback?

shieldsjared: Sorry, was moving over the past week and just getting back to work. Will get this wrapped up soon :)

shieldsjared: @hanzei - Updated it based on the plan you outlined. If teamID is empty, then the method will post to all demo channels (e.g. for each team).

hanzei (approved):
Awesome work! Thank you so much @shieldsjared :rocket:

hanzei:
@DHaussermann This PR should be tested alongside https://github.com/mattermost/mattermost-server/pull/15765

PR #149, opened 2021-10-12
Matched FileWillBeUploaded Signature to Defined in Interface

### LINKED ISSUE #125: FileWillBeUploaded does not implement the server hook
The `FileWillBeUploaded` hook in [file_hooks.go](https://github.com/mattermost/mattermost-plugin-demo/blob/master/server/file_hooks.go) is never called by the server, as it does not implement the function defined by the interface.

The plugin's function signature is:
```go
FileWillBeUploaded(c *plugin.Context, fileInfo *model.FileInfo, reader bytes.Reader, buf *bytes.Buffer)
```

but the one defined in [the interface](https://github.com/mattermost/mattermost-server/blob/master/plugin/hooks.go#L193) is:

```go
FileWillBeUploaded(c *Context, info *model.FileInfo, file io.Reader, output io.Writer) (*model.FileInfo, string)
```

We need to change the signature to match the one defined by the interface and check that it actually gets called when a file is uploaded.

adithyaakrishna:
Signed-off-by: Adithya Krishna <aadithya794@gmail.com>

#### Summary
Updates FileWillBeUploaded function signature to be similar to what's defined in the interface

#### Ticket Link
This PR Fixes #125

### DIFF
--- a/server/file_hooks.go
+++ b/server/file_hooks.go
@@ -16,7 +16,7 @@ func (p *Plugin) FileWillBeUploaded(c *plugin.Context, fileInfo *model.FileInfo,
 	configuration := p.getConfiguration()
 
 	if configuration.disabled {
-		return
+		return nil, "Configuration is disabled"
 	}
 
 	teams, err := p.API.GetTeams()
@@ -25,7 +25,7 @@ func (p *Plugin) FileWillBeUploaded(c *plugin.Context, fileInfo *model.FileInfo,
 			"Failed to query teams FileWillBeUploaded",
 			"error", err.Error(),
 		)
-		return 
+		return nil, "Failed to query teams"
 	}
 
 	if reader.Size() == 0 {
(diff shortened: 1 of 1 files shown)


server/file_hooks.go
```diff
@@ -46,4 +46,5 @@ func (p *Plugin) FileWillBeUploaded(c *plugin.Context, fileInfo *model.FileInfo,
 			)
 		}
 	}
+	return nil, "" // <--- This is the new return line
```
agarciamontoro, line 49: Can we remove the comment in this line? I wrote it in my example simply as a guide.

agarciamontoro (changes requested):
Thanks, @adithyaakrishna! Just one more thing I commented below and we're good to go :rocket:

agarciamontoro:
Hi, @adithyaakrishna! Is there anything we can do to help you with this?

adithyaakrishna:
@agarciamontoro Oops! I had missed that one, I am working on it rn and will send a commit regarding this soon :)

server/file_hooks.go
adithyaakrishna: Thanks for pointing that out :)

hanzei (approved):
Nice :+1:

adithyaakrishna:
Yayyy 🎉

PR #161, opened 2023-06-05
[MM-52924] Add `ConfigurationWillBeSaved` hook example

streamer45:
#### Summary

Adding an example for the new `ConfigurationWillBeSaved` hook. 

Still not entirely sure how we deal with new hooks on this side given the server dependency is rather out of date. Are we expecting users to manually update it? Otherwise I don't see how new hooks would ever work given the plugin gets compiled without the required changes.

#### Server PR

https://github.com/mattermost/mattermost/pull/23567

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-52924


server/configuration.go
```diff
@@ -231,6 +231,84 @@ func (p *Plugin) OnConfigurationChange() error {
...
+	js, err := json.Marshal(configData)
+	if err != nil {
+		p.API.LogError(
+			"Failed to marshal config data ConfigurationWillBeSaved",
+			"error", err.Error(),
+		)
+		return nil, nil
+	}
+
+	if err := json.Unmarshal(js, &cfg); err != nil {
```
hanzei, line 273: Nit pick: Is is worth allocating another variable here to make it easier to compare new and old config?

hanzei (approved):
LGTM :+1:

Thanks for taking care of the example, Claudio

server/configuration.go
streamer45: Sure, although the example doesn't really do any comparison with the previous one.

cpoile (approved):
> Still not entirely sure how we deal with new hooks on this side given the server dependency is rather out of date. Are we expecting users to manually update it? Otherwise I don't see how new hooks would ever work given the plugin gets compiled without the required changes.

I'm not really sure what you mean by this, can you reword it?

streamer45:
> > Still not entirely sure how we deal with new hooks on this side given the server dependency is rather out of date. Are we expecting users to manually update it? Otherwise I don't see how new hooks would ever work given the plugin gets compiled without the required changes.
> 
> I'm not really sure what you mean by this, can you reword it?

@cpoile As is if you compile the demo plugin it won't run the hook because it's pointing to a very out of date server version so there isn't much to demo, code is mostly documentation at this point. We need to update everything, which also means the `plugin-api` dependency so not straighforward.

cpoile:
> > > Still not entirely sure how we deal with new hooks on this side given the server dependency is rather out of date. Are we expecting users to manually update it? Otherwise I don't see how new hooks would ever work given the plugin gets compiled without the required changes.
> > 
> > 
> > I'm not really sure what you mean by this, can you reword it?
> 
> @cpoile As is if you compile the demo plugin it won't run the hook because it's pointing to a very out of date server version so there isn't much to demo, code is mostly documentation at this point. We need to update everything, which also means the `plugin-api` dependency so not straighforward.

Ah, ok, gotcha. Have we decided what we're doing with plugin-api @hanzei ? Would it be worth upgrading it, and the demo? (this would be a good chance)

PR #172, opened 2023-12-04
Add user settings registry

larkox:
#### Summary
Add user settings functionality to the demo plugin

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-55755

#### Related to:
https://github.com/mattermost/mattermost/pull/25517


webapp/package.json
larkox, line 1: For some reason I wasn't able to build the webapp part of the plugin, so I took the liberty to bump the dependencies.

marianunez (changes requested):
Can we add the documentation in the webapp Readme of how this new registry would look like?  here -> https://github.com/mattermost/mattermost-plugin-demo/tree/master/webapp#main-menu-action

marianunez (approved):
Thanks @larkox! Do we need to bump the min server version for support of this new registry? https://github.com/mattermost/mattermost-plugin-demo/blob/308bdd639fb46a97ab0487a69bf4260b6ae16127/plugin.json#L9C6-L9C24

larkox:
@marianunez Not really. The code checks whether the registry exist, and use it if it exist, and doesn't use it if not. Therefore... the plugin can work with previous versions (it won't show the settings, thought).

yasserfaraazkhan (approved):
able to save options
image

PR #183, opened 2025-08-04
Adds MultiForm to demo plugin for testing/demoing

sbishel:
#### Summary
Adds MultiForm to demo plugin for testing/demoing

see https://github.com/mattermost/mattermost/pull/33076

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-65163

### DIFF
--- a/server/command_hooks.go
+++ b/server/command_hooks.go
@@ -39,6 +39,7 @@ const (
 		"- `/dialog relative-callback-url` - Open an Interactive Dialog with relative callback URL. Once submitted, user's action is posted back into a channel.\n" +
 		"- `/dialog introduction-text` - Open an Interactive Dialog with optional introduction text. Once submitted, user's action is posted back into a channel.\n" +
 		"- `/dialog dynamic-select` - Open an Interactive Dialog with dynamic select fields. Once submitted, user-entered input is posted back into a channel.\n" +
+		"- `/dialog multi-select` - Open an Interactive Dialog with multi-select fields. Once submitted, user-entered input is posted back into a channel.\n" +
 		"- `/dialog error` - Open an Interactive Dialog which always returns an general error.\n" +
 		"- `/dialog error-no-elements` - Open an Interactive Dialog with no elements which always returns an general error.\n" +
 		"- `/dialog field-refresh` - Open an Interactive Dialog with field refresh functionality.\n" +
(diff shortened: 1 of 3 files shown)


server/http_hooks.go
```diff
@@ -412,25 +415,204 @@ func (p *Plugin) handleDynamicArgTest(w http.ResponseWriter, r *http.Request) {
...
+
+	rootPost, appErr := p.API.CreatePost(&model.Post{
+		UserId:    p.botID,
+		ChannelId: request.ChannelId,
+		Message:   msg,
+	})
+	if appErr != nil {
+		p.API.LogError("Failed to post field refresh dialog result", "err", appErr.Error())
+		return
+	}
```
sbishel, line 492: Should check isCancelled and update or remove message

server/command_hooks.go
```diff
@@ -36,6 +36,8 @@ const (
 		"- `/dialog introduction-text` - Open an Interactive Dialog with optional introduction text. Once submitted, user's action is posted back into a channel.\n" +
 		"- `/dialog error` - Open an Interactive Dialog which always returns an general error.\n" +
 		"- `/dialog error-no-elements` - Open an Interactive Dialog with no elements which always returns an general error.\n" +
+		"- `/dialog field-refresh` - **NEW:** Open an Interactive Dialog with field refresh functionality (requires Apps Form feature flag).\n" +
+		"- `/dialog multistep` - **NEW:** Open a multi-step Interactive Dialog demonstrating form refresh on submit (requires Apps Form feature flag).\n" +
```
hanzei, line 40: Interesting that the feature flag is needed despite your changes in https://github.com/mattermost/mattermost/pull/33076. In wonder if we can drop that requirement.

server/dialog_samples.go
hanzei, line 1: Should we move the other dialogs into this file as well?

hanzei, line 1: Neat example!

hanzei (changes requested):
Thanks for keeping the demo plugin up to date!

I tried running the changes locally, but I did run into go mod issues. Does `github.com/mattermost/mattermost/server/public` need an update?

server/dialog_samples.go
sbishel: done

sbishel:
> I tried running the changes locally, but I did run into go mod issues. Does `github.com/mattermost/mattermost/server/public` need an update?

Yes, it does need the latest changes. I updated and once v1.20 is released, we should update to that version.

All of these features came into the server over the last couple versions...I think, may-be I should add the server versions to the autocomplete text. I'll create a ticket for that.

PR #186, opened 2025-08-14
refactor: pluginctl v0.1.5-0.20250814142442-19c417d53e1d

fmartingr:
#### Summary

Example PR on how to migrate a plugin to `pluginctl`.

1. Add `webapp.config.js` and `babel.config.js` to the manifest `props.pluginctl.ignore_assets`.
2. Remove `build/setup.mk`, `build/manifest` & `build/pluginctl`.
3. Run `pluginctl updateassets`.

Migration guide: https://github.com/mattermost/pluginctl/pull/2

Closes: #180


plugin.go
hanzei, line 1: This goes back to the changes in https://github.com/mattermost/mattermost-plugin-starter-template/pull/192. With that PR, we auto-generate the manifest every time we build the bundle, because the `version` field now depends on the git history. The demo plugin never adopted the new pattern.

The changes in this PR are fine.

.golangci.yml
hanzei, line 1: 👍

hanzei (changes requested):
I'm a bit hesitant to migrate this repo over until the discussion about pluginctl is resolved. I'm blocking until then.

PR #187, opened 2025-09-28
fix makefile to correctly create plugin bundle archive

xangcastle:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -228,7 +228,7 @@ ifneq ($(HAS_WEBAPP),)
 	mkdir -p dist/$(PLUGIN_ID)/webapp
 	cp -r webapp/dist dist/$(PLUGIN_ID)/webapp/
 endif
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -228,7 +228,7 @@ ifneq ($(HAS_WEBAPP),)
 	mkdir -p dist/$(PLUGIN_ID)/webapp
 	cp -r webapp/dist dist/$(PLUGIN_ID)/webapp/
 endif
-	cd dist && tar -cvzf $(BUNDLE_NAME) $(PLUGIN_ID)
+	cd dist/$(PLUGIN_ID) && tar -cvzf ../$(BUNDLE_NAME) .
```
hanzei, line 231: I do wonder what prompted this change. The MM server accepts both the previous and the new formats. 

Why the change?

hanzei (changes requested):

PR #204, opened 2026-03-13
Add channel settings tab demo

nickmisasi:
## Summary
- Register a `ChannelSettingsTab` pluggable to smoke-test the new channel settings tab plugin API
- Add a minimal `ChannelSettingsSmokeTest` component that renders channel info, a dirty-state input, and a tab-switch error banner
- Make `GetTeams()` calls non-fatal in `ensureDemoUser`, `ensureDemoChannels`, `OnActivate`, and `OnDeactivate` so the plugin can start even when the Teams table has NULL string columns (core server bug where `CompanyName` is NULL) 
Needs https://github.com/mattermost/mattermost/pull/35591

## Test plan
- [ ] Build and deploy the plugin to a Mattermost instance with channel settings pluggable support
- [ ] Open Channel Settings on a public/private channel and confirm the **Smoke Test** tab appears
- [ ] Click the tab and verify it renders the channel display name, name, and ID
- [ ] Type into the input to mark the tab dirty, then try switching tabs to confirm the unsaved-change warning
- [ ] Verify the plugin starts cleanly even if `GetTeams()` returns an error (check server logs for warn-level messages instead of crashes)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


server/activate_hooks.go
```diff
@@ -38,19 +38,19 @@ func (p *Plugin) OnActivate() error {
...
-		_, ok := configuration.demoChannelIDs[team.Id]
-		if !ok {
-			p.API.LogWarn("No demo channel id for team", "team", team.Id)
-			continue
-		}
-
-		msg := fmt.Sprintf("OnActivate: %s", manifest.Id)
-		if err := p.postPluginMessage(team.Id, msg); err != nil {
-			return errors.Wrap(err, "failed to post OnActivate message")
+		p.API.LogWarn("Failed to query teams OnActivate, skipping activation messages", "error", err.Error())
```
hanzei, line 41: It's strange that `GetTeams` fails. I wonder if we should still fail loudly as that makes things easier for QA to debug.

webapp/src/components/channel_settings_smoke_test.jsx
hanzei, line 1: I would love to see a typescript file. not a blocker

hanzei (commented):
I'm not sure I agree with the non-fatal changes

PR #210, opened 2026-05-04
Add MCP demo tools

nickmisasi:
## Summary
- Adds demo MCP server wiring and tools using `mcphelper`.
- Registers/unregisters MCP best-effort during plugin lifecycle.
- Depends on source PR: https://github.com/mattermost/mattermost-plugin-agents/pull/678

## Test plan
- `go test ./server`

Made with [Cursor](https://cursor.com)


nickmisasi:
@hanzei gentle ping

go.mod
```diff
@@ -34,28 +37,36 @@ require (
...
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20260504160031-60b97b32f348 // indirect
+	google.golang.org/grpc v1.81.0 // indirect
 	google.golang.org/protobuf v1.36.11 // indirect
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+// Match pluginmcp's server/public pin for helper compatibility.
+replace github.com/mattermost/mattermost/server/public => github.com/mattermost/mattermost/server/public v0.3.1-0.20260402155910-d9d71af83e3f
```
hanzei: It's not clear to me, why we need that. Is there new code in the `public` packet?

server/activate_hooks.go
```diff
@@ -10,11 +10,33 @@ import (
...
 )
 
+var (
+	onActivateCore = func(p *Plugin) error {
+		return p.onActivateCore()
+	}
+	onDeactivateCore = func(p *Plugin) error {
+		return p.onDeactivateCore()
+	}
+)
```
hanzei: That seems quite hacky.

server/mcp.go
```diff
@@ -0,0 +1,99 @@
...
+		ExposeExternal: true,
+		Version:        manifest.Version,
+	})
+
+	p.registerMCPTools(server)
+	p.mcpServer = server
+	return nil
+}
+
+func (p *Plugin) registerMCPServerBestEffort() {
```
hanzei, line 46: Why does the method log an error instead of returning it? Do we expect it to fail often?

server/mcp.go
```diff
@@ -0,0 +1,99 @@
...
+
+	if err := mcpUnregister(server); err != nil {
+		p.API.LogWarn("MCP unregister failed; continuing plugin shutdown", "err", err.Error())
+	}
+}
+
+func (p *Plugin) serveMCPIfMatch(w http.ResponseWriter, r *http.Request) bool {
+	if r.URL.Path != mcpBasePath && !strings.HasPrefix(r.URL.Path, mcpBasePath+"/") {
+		return false
+	}
```
hanzei: 1/5 we could just add a router instead of writing this custom code

server/mcp_tools.go
```diff
@@ -0,0 +1,78 @@
...
+	Sum int `json:"sum" jsonschema:"Sum of a and b"`
+}
+
+type GetUserDisplayNameArgs struct{}
+
+type GetUserDisplayNameOutput struct {
+	UserID      string `json:"user_id" jsonschema:"Mattermost user ID of the caller"`
+	Username    string `json:"username" jsonschema:"Username of the caller"`
+	DisplayName string `json:"display_name" jsonschema:"Full-name display name of the caller, falling back to username"`
+}
```
hanzei, line 35: Neat 💯

hanzei (changes requested):
This is awesome! Looking forward to seeing more plugins exposed via MCP. Boards is next?

I'm not too happy about the testing pattern. Testing with mocks has caused a lot of issues in the past and the coverage is not great. I wonder if we should make use of https://github.com/mattermost/mattermost/pull/35643 and test against a real deployment of a plugin. Maybe with a mock upstream server.

server/mcp.go
nickmisasi: I’m intentionally keeping MCP registration best-effort here. This plugin should still activate as a demo plugin even when the Agents plugin/MCP registration path is unavailable, and the warning log makes that degraded state visible.

nickmisasi:
@hanzei I agree real plugin deployment coverage would be stronger, but I’d prefer not to add a Docker/testcontainers integration suite to this demo PR while the shared testhelper is still landing. It feels a bit overkill for a demo plugin example implementation. We have tests in the Agents plugin, and there are real implementations happening in both the Playbooks and Cloud plugins right now - those would be better spots for that.

server/activate_hooks_test.go
hanzei, line 1: I'm not sure if the tests in this file add much value. The mostly test the logging.

hanzei: If we remove them, we can also cleanup the `mcpRegister` workaround

go.mod
```diff
@@ -1,14 +1,15 @@
 module github.com/mattermost/mattermost-plugin-demo
 
-go 1.25.8
+go 1.26.2
 
 require (
 	github.com/gorilla/mux v1.8.1
-	github.com/mattermost/mattermost/server/public v0.3.0
+	github.com/mattermost/mattermost-plugin-agents v0.0.0-20260508170706-18844c90064e
+	github.com/mattermost/mattermost/server/public v0.3.1-0.20260402155910-d9d71af83e3f
```
hanzei, line 8: Is there a tagged release we can use? If not, should we tag one?

hanzei (changes requested):
> It feels a bit overkill for a demo plugin example implementation.

I'm not sure if I agree. The demo plugins is often used an the example implementation - both from humans and from LLMs. I'm leaning towards providing a "good", golden path implementation here that other projects can simply follow. Extracting that "golden path" from other projects like calls is not straightforward as these projects are often tainted with other requirements.

Would it be an option to file a follow up ticket and work on that once the test containers are in `master`?

server/activate_hooks_test.go
nickmisasi: Removed

go.mod
nickmisasi: There's a go module issue with the agents plugin being on v2. I've pointed at the commit sha for the v2.1.0 tag for now

nickmisasi:
>I'm leaning towards providing a "good", golden path implementation here that other projects can simply follow. Extracting that "golden path" from other projects like calls is not straightforward as these projects are often tainted with other requirements.

@hanzei I think this is a great intention in theory, but it sort of falls apart when you consider that this repository as it stands has virtually no tests at all (this PR has added one server test file, now totals 3) :P  

An E2E deployment test here would mostly test another plugin/service’s MCP infrastructure rather than this demo plugin’s code. We should keep this PR scoped to unit/HTTP-level coverage for the example wiring, and rely on the Agents/pluginmcp project for full cross-plugin integration coverage. IMO that _is_ the golden path for a plugin implementing this.

Keep in mind there's more to this than just having a Mattermost testcontainer - you need to deploy and setup the Agents plugin, and then either completely mock out (difficult, fragile) an upstream LLM provider, or wire in a real one (expensive, also fragile because they're slow). Simply having the testcontainers for mattermost merged isn't going to be exhaustive enough

hanzei (approved):

PR #219, opened 2026-08-21
Add mmBlocks utils functions for e2e

larkox:
#### Summary
Add MM Blocks functionalities for e2e testing

#### Ticket Link
NONE


server/command_hooks.go
```diff
@@ -23,6 +23,7 @@ const (
 	commandTriggerAutocompleteTest  = "autocomplete_test"
 	commandTriggerToast             = "toast"
 	commandTriggerInlineAction      = "inline_action"
+	commandTriggerMmBlocks          = "mm_blocks"
```
hanzei, line 26: I think we dropped the term `MM Blocks` from the docs, right? Should we drop it from this code as well?

server/mm_blocks_command.go
```diff
@@ -0,0 +1,295 @@
...
+						map[string]any{"text": "Alpha", "value": "opt_alpha"},
+						map[string]any{"text": "Beta", "value": "opt_beta"},
+						map[string]any{"text": "Gamma", "value": "opt_gamma"},
+					},
+				},
+			},
+			map[string]any{
+				"demo_mm_blocks_select": mmBlocksPostAction("/mm_blocks_integration_static_select", map[string]any{}),
+			},
+		)
```
hanzei, line 147: UX issue: When the user selects an option and then clicks on the dialog again, the picker goes back to neutral state. I think it should highlight which option was picked in the first run.

Image

Image

server/mm_blocks_command.go
```diff
@@ -0,0 +1,295 @@
...
+					"text":      "Submit form",
+					"style":     "primary",
+					"subtype":   "submit",
+					"action_id": "demo_mm_blocks_form",
+				},
+			},
+			map[string]any{
+				"demo_mm_blocks_form": mmBlocksPostAction("/mm_blocks_integration_echo_form_values", map[string]any{}),
+			},
+		)
```
hanzei, line 189: This does not seem to work. Am I missing something?

Image

server/mm_blocks_command.go
```diff
@@ -0,0 +1,295 @@
...
+					"style":     "primary",
+					"subtype":   "submit",
+					"action_id": "demo_mm_blocks_form",
+				},
+			},
+			map[string]any{
+				"demo_mm_blocks_lookup": mmBlocksPostAction("/mm_blocks_integration_lookup", map[string]any{}),
+				"demo_mm_blocks_form":   mmBlocksPostAction("/mm_blocks_integration_echo_form_values", map[string]any{}),
+			},
+		)
```
hanzei, line 214: Same here

server/mm_blocks_command.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	switch command {
+	case "help":
+		return &model.CommandResponse{
+			ResponseType: model.CommandResponseTypeEphemeral,
+			Text:         commandMmBlocksHelp,
+		}
+	case "", "demo":
+		return p.postMmBlocksDemo(args)
```
hanzei, line 85: It's a bit odd to have both sub demo and a main demo, if they also overlap. I suggest we 

a) make the two demos not overlap or
b) drop the main demo, i.e. `/mm_blocks`

server/mm_blocks_command.go
```diff
@@ -0,0 +1,295 @@
...
+	post := &model.Post{
+		ChannelId: args.ChannelId,
+		RootId:    args.RootId,
+		UserId:    p.botID,
+		Message:   message,
+		Props: model.StringInterface{
+			"mm_blocks":         blocks,
+			"mm_blocks_actions": actions,
+		},
+	}
```
hanzei, line 298: Wasn't there support to sections of blocks? Should we exercise that here?

hanzei (changes requested):
Great to see the Blocks comming to the demo plugin 🎉 

Below are my high level comments. Let's resolve them first before I dive deep into the code.

server/command_hooks.go
larkox: We need to keep a differentiation between "blocks" (used by Slack's block kit) and our own blocks.

server/mm_blocks_command.go
larkox: Out of the scope of this PR, but will take a look for the PR in the main repo.

server/mm_blocks_command.go
larkox: I have to double check, but in order for these to work as they should, you need these PRS:
https://github.com/mattermost/mattermost/pull/37767
https://github.com/mattermost/mattermost-mobile/pull/10024

server/mm_blocks_command.go
larkox: I can update this so the main goes to help, instead of demo.

server/mm_blocks_command.go
larkox: The idea was to port only what was needed for e2e tests. Some features (like collapsing blocks) were already being tested in a different way without the need of an external service, and that is why this was not here yet. But you are right this is a good idea.

I had already a set of fixtures in the server, so I ported them here under a "examples" command.

In particular, take a look to the "incident response" one, which tries to use as much as possible of blocks.

hanzei (changes requested):
Blocking on the naming pattern. If `Mattermost Blocks` is what the docs are talking about, `Blocks` is fine in a Mattermost context. `mm_block` opens a discussion we closed already.

